PR #2, opened 2022-09-13
Added API tests, updated namespace

3coins:
- Added unit tests for api handlers
- Changed namespace of all apis to include "scheduler"
- Made some pre-commit fixes, removed mypy

### DIFF
--- a/jupyter_scheduler/tests/test_handlers.py
+++ b/jupyter_scheduler/tests/test_handlers.py
@@ -1,13 +1,262 @@
 import json
+from contextlib import nullcontext
+from unittest.mock import patch
+
+import pytest
+from tornado.httpclient import HTTPClientError
+
+from jupyter_scheduler.handlers import compute_sort_model
+from jupyter_scheduler.models import (
+    CountJobsQuery,
+    DescribeJob,
+    ListJobsQuery,
+    ListJobsResponse,
+    SortDirection,
+    SortField,
+    Status,
+)
+from jupyter_scheduler.tests.utils import expected_http_error
 
 
-async def test_get_example(jp_fetch):
-    # When
-    response = await jp_fetch("jupyter-scheduler", "get_example")
+@pytest.mark.parametrize(
+    "job_id,payload",
+    [
+        (
+            "542e0fac-1274-4a78-8340-a850bdb559c8",
+            {
+                "input_uri": "notebook_a.ipynb",
+                "output_prefix": "outputs",
+                "idempotency_token": "",
+                "runtime_environment_name": "",
+            },
+        ),
+        (
+            "4c6cd4e0-49ce-4b58-843d-2fa02f7468b1",
+            {
+                "input_uri": "notebook_b.ipynb",
+                "output_prefix": "scheduled_outputs",
+                "idempotency_token": "",
+                "runtime_environment_name": "",
+                "name": "Scheduled Notebook B",
+                "job_definition_id": "7790f93c-4c2c-41b2-9085-daa93915d81c",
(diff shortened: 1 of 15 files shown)


3coins:
Ignore the failures for now. I will tackle the CI setup once all the existing code is pushed to Github.

jupyter_scheduler/tests/test_handlers.py
```diff
@@ -1,13 +1,262 @@
...
+    ],
+)
+async def test_post_scheduled_jobs(jp_fetch, job_id, payload):
+    with patch("jupyter_scheduler.orm.uuid4") as mock_uuid:
+        mock_uuid.return_value = job_id
+        response = await jp_fetch("scheduler", "jobs", method="POST", body=json.dumps(payload))
+
+        assert response.code == 200
+        body = json.loads(response.body)
+        assert body["job_id"]
```
dlqqq, line 54: Shouldn't this be:

```
assert body["job_id"] == job_id
```

jupyter_scheduler/tests/test_handlers.py
```diff
@@ -1,13 +1,262 @@
...
+async def test_batch_delete(jp_fetch):
+    with patch("jupyter_scheduler.scheduler.Scheduler.delete_job") as mock_delete_job:
+        job_id = "542e0fac-1274-4a78-8340-a850bdb559c8"
+        response = await jp_fetch(
+            "scheduler", "batch", "jobs", method="DELETE", params={"job_id": job_id}
+        )
+
+        mock_delete_job.assert_called_once_with(job_id)
+        assert response.code == 204
+
```
dlqqq, line 202: These tests (patch/delete) are not meaningful as they should be. You are only asserting that the response code is OK, and not verifying their intended side effect of updating/deleting jobs. Ideally, each of your existing assertions should be followed by a get request that actually verifies the corresponding job was updated/deleted.

dlqqq (changes requested):

jupyter_scheduler/tests/test_handlers.py
3coins: My thought process was adding testing in layers; these tests covering the API layer, only test the api portion making sure they call the correct scheduler method (mocking the scheduler methods here) with expected parameters. There are not a lot of exception cases here, but in future I expect some more tests to cover those scenarios.
I intend to write addition tests that will cover the scheduler portion of the code which will test actual persistence of jobs, but will add those in a separate test file.

jupyter_scheduler/tests/test_handlers.py
3coins: Good catch. Thanks for pointing this out.

jupyter_scheduler/tests/test_handlers.py
dlqqq: That's fine if you plan on adding more tests in the future. I do have a high-level comment: while I understand that your testing strategy conforms more strictly to proper "unit testing", I just don't see the value in separating the handler and business logic tests. IMO, it's just easier to write tests that cover both, while also giving you more confidence that your server actually works.

dlqqq (approved):

jupyter_scheduler/tests/test_handlers.py
3coins: @dlqqq Thanks for the suggestions. I agree there is value in integration tests that cover the full service function, but as we are evolving and changing things frequently right now, having unit tests keeps us more pliable at the moment. 
I am also following what some other jupyterlab extensions are doing. For example, here are the tests for jupyterlab-git extension.
https://github.com/jupyterlab/jupyterlab-git/blob/master/jupyterlab_git/tests/test_handlers.py

PR #12, opened 2022-09-16
CollapsiblePanel

andrii-i:
This PR introduces reusable CollapsiblePanel UI element and styling for it

Preview:
children

References:
- [Material UI Accordion](https://mui.com/material-ui/react-accordion/)
- Material React Accordion [implementation](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Accordion/Accordion.js)

Please note that component is not currently used at Create Job page nor anywhere in the repo (pages to record GIFs were created locally without committing anything). To use component inside the package:

- import CollapsiblePanel from /src/components
- instantiate a component by nesting panel content in the element and providing props:
  -   expanded?: boolean;
  -   title: string;

Use example:
Screen Shot 2022-09-19 at 5 21 20 PM


src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,43 @@
+import { ToolbarButtonComponent } from '@jupyterlab/apputils';
+import React, { useState } from 'react';
+import { caretDownIcon, caretRightIcon } from '@jupyterlab/ui-components';
+
+export interface ICollapsiblePanelProps {
+  expandedByDefault?: boolean;
```
3coins, line 6: **nit**: Move optional attributes after required ones.

src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <span className="jp-collapsible-panel-title" onClick={onControlClick}>
+          {props.title}
+        </span>
+      </div>
+      <div
+        className={
+          'jp-collapsible-panel-content' + (expanded ? ' expanded' : '')
+        }
+      >
+        {expanded && <div>{props.content}</div>}
```
3coins, line 39: Wouldn't this remove the content rather than hiding it? Is that the intended effect here?

src/components/collapsible-panel.tsx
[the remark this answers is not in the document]
jweill-aws: This would give us `expanded` in both props and state, though

src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <div
+        className={
+          'jp-collapsible-panel-summary' + (expanded ? ' expanded' : '')
+        }
+      >
+        <ToolbarButtonComponent
+          className="jp-collapsible-panel-icon-toolbar-button"
+          icon={expanded ? caretDownIcon : caretRightIcon}
+          onClick={onControlClick}
+        />
```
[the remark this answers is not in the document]
jweill-aws: I originally used `Button` for the action icons in each row, but even the "minimal" styling on buttons drove me away from `Button` and towards `ToolbarButtonComponent`. Displaying the icon inline could work as an alternative, though, since we don't intend to use it as a button by itself.

src/components/collapsible-panel.tsx
jweill-aws: Agreed — if the content isn't visible, it might be better to leave it in the DOM so that other functions can modify it.

style/base.css
```diff
@@ -231,3 +231,34 @@ ul.jp-notebook-job-output-formats-options {
...
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+}
+
+.jp-collapsible-panel-icon-toolbar-button {
+  cursor: default;
+  font-weight: bold;
+  background-color: unset !important;
+  color: unset !important;
```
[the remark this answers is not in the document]
jweill-aws: Providing a more [specific](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) rule would be better here.

src/components/collapsible-panel.tsx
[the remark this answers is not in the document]
andrii-i: As of now this prop is only used to provide initial state of the collapsible component only, it doesn’t control the state of collapsible component “in real time.”  Taking this in account, do you think name change would be appropriate? Also if you think there needs to be ability to make component “controlled” / add actual “expanded?” prop, it’s also a valid point

andrii-i: Done

style/base.css
```diff
@@ -231,3 +231,34 @@ ul.jp-notebook-job-output-formats-options {
...
+}
+
+.jp-collapsible-panel-title {
+  font-size: var(--jp-ui-font-size2);
+  font-weight: bold;
```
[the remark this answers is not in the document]
andrii-i: Done, var works

src/components/collapsible-panel.tsx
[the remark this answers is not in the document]
andrii-i: Renamed prop to `expanded ` to be less verbose. Additional clarity could be provided in docs when we'll have them as pointed out by @3coins

src/components/collapsible-panel.tsx
andrii-i: That's a good point, thank you both for pointing this out. Changed to hiding content when panel is closed via css `display: none;`

style/base.css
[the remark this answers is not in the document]
andrii-i: Thanks for pointing me to Specificity @jweill-aws, that's very useful

src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,43 @@
...
+  title: JSX.Element | string;
+  content: JSX.Element | string;
+  expanded?: boolean;
+}
+
+export function CollapsiblePanel(props: ICollapsiblePanelProps): JSX.Element {
+  const [expanded, setExpanded] = useState(props.expanded ?? false);
+
+  const handleChange = () => {
+    setExpanded(!expanded);
```
jweill-aws, line 15: Is this function used anywhere else? If not, it's probably easier to include its one line in the `onClick` argument's value.

andrii-i: Its used on twice / on 2 elements: summary icon and summary title. Based on this decided to put click logic into single function as a single source of truth

style/base.css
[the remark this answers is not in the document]
andrii-i: Got rid of the need to override styles by using the icon directly

src/components/collapsible-panel.tsx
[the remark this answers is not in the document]
andrii-i: Got rid of `ToolbarButtonComponent` by using `caretDownIcon/caretRightIcon.react` to render the icon inline directly

style/base.css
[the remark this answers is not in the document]
andrii-i: It looks different with JL theme var for bold text though:

jl var for bold

andrii-i:
All comments addressed. Please also find most current preview below (no padding on collapsible content). It looks a bit different because Jupyterlab var is used for bold text of CollapsiblePanel title:

jl var for bold

andrii-i:
Please also note that component is not currently used at Create Job page nor anywhere in the repo (pages to record GIFs were created locally without committing anything).

src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,44 @@
...
+  };
+
+  return (
+    <div className={'jp-collapsible-panel' + (expanded ? ' expanded' : '')}>
+      <div
```
3coins, line 19: What is the purpose of having a summary container? The way I see this component is just 2 sections, 1) header and 2) body. Header should include the icon and the title, since clicking on either should expand/close. Here is what I propose to simplify this component.

``` html
<div class='jp-collapsible-panel'>
    <div class='jp-collapsible-panel-header' onClick={handleChange}>
        <icon></icon>
        <span class='jp-collapsible-panel-title'>{title}</span>
    </div>
    <div class='jp-collapsible-panel-body'>
        {content}
    </div>
</div>
```

src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,44 @@
...
+        className={
+          'jp-collapsible-panel-summary' + (expanded ? ' expanded' : '')
+        }
+      >
+        <div onClick={handleChange} className="jp-collapsible-panel-button">
+          {expanded ? (
+            <caretDownIcon.react container={document.createElement('SPAN')} />
+          ) : (
+            <caretRightIcon.react container={document.createElement('SPAN')} />
```
3coins, line 28: Is the container property required here?

andrii-i: used tag="span" instead of container property as done in other places where [.react is used in jupyterlab codebase](https://github.com/search?q=org%3Ajupyterlab+Icon.react&type=code). Serves the same purpose of providing container to icon. If not provided, defaults to <div>

andrii-i: Figured out how to remove a container

src/components/collapsible-panel.tsx
andrii-i: Good point, simplified as outlined above and also by using <h2> instead of <div>

src/components/collapsible-panel.tsx
andrii-i: After simplifying the number of containers used, changed to inline function on `onClick` as suggested

style/base.css
```diff
@@ -231,3 +231,29 @@ ul.jp-notebook-job-output-formats-options {
 .jp-notebook-job-navigation-tab.active {
   border-block-end: 2px solid var(--jp-brand-color1);
 }
+
+
+.jp-collapsible-panel {
```
ellisonbg, line 236: Let's use the CSS class naming convention of `.jp-jobs-<Component>` so `jp-jobs-CollapsiblePanel`.

style/base.css
```diff
@@ -231,3 +231,29 @@ ul.jp-notebook-job-output-formats-options {
 .jp-notebook-job-navigation-tab.active {
   border-block-end: 2px solid var(--jp-brand-color1);
 }
+
```
ellisonbg, line 234: Let's move the CSS for this component into a separate CSS file that is imported into base.css, following the practice in #13

src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+import { caretDownIcon, caretRightIcon } from '@jupyterlab/ui-components';
+
+export interface ICollapsiblePanelProps {
+  title: JSX.Element | string;
+  content: JSX.Element | string;
```
ellisonbg, line 6: Does it make sense for us to use `children` for this? Maybe review how other UI libraries handle this.

src/components/collapsible-panel.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+import { caretDownIcon, caretRightIcon } from '@jupyterlab/ui-components';
+
+export interface ICollapsiblePanelProps {
+  title: JSX.Element | string;
```
ellisonbg, line 5: Do we want to allow the title to be any JSX element or should we constraint it to be text? That would make styling easier.

style/base.css
```diff
@@ -231,3 +231,29 @@ ul.jp-notebook-job-output-formats-options {
...
+
+.jp-collapsible-panel {
+  inline-size: 100%;
+}
+
+.jp-collapsible-panel-header {
```
ellisonbg, line 240: If we can constraint the title to be text, let's size the vertical height of this using the `--jp-size` variables in #13

ellisonbg (changes requested):

src/components/collapsible-panel.tsx
andrii-i: As of now, our component is minimalistic (e.g. we don't provide options for icon/control) so I think restricting title to just string makes sense

src/components/collapsible-panel.tsx
andrii-i: Material Accordion [uses](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Accordion/Accordion.js#L151) `children` for both Title/Summary and Content where first element of `children` is Title/Summary and everything else is Content by convention. I think it's a straightforward approach (going to implement it now). But [based on other comment](https://github.com/jupyter-server/jupyter-scheduler/pull/12#discussion_r974614651), Title should still be a prop, right?

style/base.css
andrii-i: I really like the adaptable styling system you built around sizes. I need #13 merged to start using it. I can implement css variables with fallback values without #13 merged but some things like AvailableSizes type I'd need are not available

style/base.css
andrii-i: Done

style/base.css
andrii-i: Done

src/components/collapsible-panel.tsx
andrii-i: Component is now using `children` for Content and separate `props.title` for Title (with `string` type enforced)

andrii-i:
Component updated to use composition / `children` for Content. Title is still a prop. Use example:
Screen Shot 2022-09-19 at 5 21 20 PM

3coins (approved):

PR #24, opened 2022-09-20
Adds extension point for custom environment UI

jweill-aws:
Adds an extension point based on the sample extension point repo.

In the image below, the component displays an informative message.

image


src/tokens.ts
```diff
@@ -0,0 +1,16 @@
+import { Token } from '@lumino/coreutils';
+
+namespace SchedulerExtension {
+  export interface ICustomEnvironmentProps {
+    state: any;
+    setState: any;
+  }
```
dlqqq, line 7: We should definitely make this type a bit more strict, and give developers a way of specifying the keys in their form extension. Let's say the type of the base form state is called `FormState`:

```
namespace SchedulerExtension {
  export type CustomFormState = Record<never, never>;

  export interface ICustomEnvironmentProps {
    state: FormState & Partial<CustomFormState>;
    setState: (state: FormState & Partial<CustomFormState>) => unknown;
  }
}
```

Developers should then be able to override `CustomFormState` via [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces). However, I have not tested this as of yet.

src/create-job-form.tsx
```diff
@@ -252,6 +254,9 @@ export function CreateJobForm(props: CreateJobFormProps): JSX.Element {
           formInput={formInput}
           fields={formFields}
         />
+        {props.customEnvironment &&
+          <props.customEnvironment state={state} setState={setState} />
+        }
```
dlqqq, line 259: It doesn't look like `props.customEnvironment` will ever be nullish, since our default plugin still returns a component. I'd say we define the type correctly in `CreateJobFormProps` as non-optional.

src/create-job-form.tsx
```diff
@@ -252,6 +254,9 @@ export function CreateJobForm(props: CreateJobFormProps): JSX.Element {
           formInput={formInput}
           fields={formFields}
         />
+        {props.customEnvironment &&
+          <props.customEnvironment state={state} setState={setState} />
```
awaisabir, line 258: One thing that users would most likely need is to run their own custom validation on the extended form fields. This could be run on the `onBlur` event for the custom form fields and we would update the validity state accordingly and you could disable the `Run Job` button if the state for this is `false`. 

```tsx
// in your extension
const [isFormValid, setIsFromValid] = useState(false); // default value of false

<props.customEnvironment state={state} setState={setState} isFormValid={isFormValid} updateValidity={setIsFormValid} />

// ---------------------------------------------------------------------------------------------------------

// in our custom env form (basic example)
const handleBlur = (e) => {
  const value = e.target.value;
  let isValid = false;

  if (value conforms to RegExp) { isValid = true; } 
  else { isValid = false; }

  updateValidity(isValid);
}
```

This would act as front-end validation for the UI piece.
Let me know if you have any questions or concerns regarding this ask!

src/tokens.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  export type ICustomEnvironment = React.FC<ICustomEnvironmentProps>;
+
+  export const ICustomEnvironment = new Token<ICustomEnvironment>(
+    'scheduler-extension-custom-environment:ICustomEnvironment'
```
3coins, line 12: I believe the convention for token id is 'namespace/extension:component'. Use `@jupyterlab/scheduler:ICustomEnvironment` here.

src/tokens.ts
```diff
@@ -0,0 +1,16 @@
+import { Token } from '@lumino/coreutils';
+
+namespace SchedulerExtension {
```
3coins, line 3:
```suggestion
namespace Scheduler {
```

src/tokens.ts
3coins: We should not provide arbitrary props, this should conform to the CreateModel attributes as the parent form will be passing in these props. Also, the custom form has to serialize all the extra properties (anything that doesn't fit into the top level CreateModel attributes) into "runtime_environment_parameters".

```ts
namespace Scheduler {
    export type EnvironmentParameterValue = string | number | boolean

    export interface CreateJobFormState {
        jobName: string
        inputUri: string
        outputPrefix: string
        runtimeEnvironmentName: string
        runtimeEnvironmentParameters?: { [key: string]: EnvironmentParameterValue }
        isValid: boolean
        // other attributes from api model
    }

    export interface ICustomEnvironmentProps {
        mode: 'Create' | 'View'
        state: CreateJobFormState
        setState: (state: CreateJobFormState) => void
    }
}
``` 
A few discussion points:
- We are calling this custom form as 'Advanced Options' in the UI, should we align with that naming here?
- I haven't thought about the Job Definition form, and if the same extension point will be used. If that's the case, then the state will support either type.

3coins (changes requested):

src/create-job-form.tsx
dlqqq: This is a good idea. However, race conditions are possible with a naive boolean implementation. Take the following sequence of events:

1. user selects from a dropdown, then network request is made to validate it
2. user moves to the next text field, typing in an invalid email address. `setValid(false)` gets invoked.
3. network request in 1) completes, and it returns true. `setValid(true)` gets invoked.

This sequence of events causes `valid` to be set to `true` despite the email field being invalid.

We need additional information as to which fields are causing the error, since a field's will not want to set the form state to valid if the even if its validate callback returns true. Thus, the `valid` state should be a dictionary mapping form keys to booleans. The form's valid state depends on whether all of the keys are `true`.

awaisabir: On the SM UI side we are maintaining a dict of mappings of form fields to error messages, where an empty string means no error essentially. For example:
```
  const [formErrors, setFormErrors] = useState({
    field1: '',
    field2: '',
    field3: '',
  });
```

Would it make more sense to the check for this validity when the `Run Job` action is clicked on? This way, we can run the validation on one pass, instead of whenever a form field is 'focused' out of.

jweill-aws: Note that we're moving our UI components to MUI, which supports validation. See docs: https://mui.com/material-ui/react-text-field/#validation

awaisabir: How would a consumer of this extension be able to perform their own custom validation, even if you are using Material UI?

jweill-aws: It looks like custom validation is possible by setting an MUI TextField's error property. Example here: https://stackoverflow.com/questions/69287291/text-field-validation-in-material-ui-using-error-property

dlqqq: @awaisabir 

> Would it make more sense to the check for this validity when the Run Job action is clicked on? This way, we can run the validation on one pass, instead of whenever a form field is 'focused' out of.

Not really, because then what's the point of client-side validation when the user has to "click" the submit button to see the failure (same UX as server-side validation)? Seems silly. The whole benefit of client-side validation is that we can fail fast, i.e. on blur events.

If this is a heavily desired feature (which I strongly doubt), then we can perhaps add a configuration option for this somehow. But until then, please perform field validation on blur events for consistency.

> How would a consumer of this extension be able to perform their own custom validation, even if you are using Material UI?

If you think about it, errors are really just additional form state. My idea for how to handle this is that your extension component would accept two props: `errors` and `setErrors`. `errors` is mapping between form key and its corresponding error message (as you had suggested), and `setErrors` is self-explanatory. You can then use an effect to listen to the field state.

```ts
useEffect(() => {
  if (isInvalid(props.state.key)) {
    props.setErrors((errors) => ({ ...errors, key: "something is wrong" }))
  } 
}, [props.state.key])

return (
  ...
  <TextField value={...} onChange={...} {...(props.errors.key && { error: true, helperText: props.errors.key }) />
)
```

Perhaps there's other ways of doing this (since curiously it doesn't look like `TextField` exposes a `validate` or `onValidate` callback prop), but this was my first thought.

awaisabir: @dlqqq 

> > Would it make more sense to the check for this validity when the Run Job action is clicked on? This way, we can run the validation on one pass, instead of whenever a form field is 'focused' out of.
> 
> Not really, because then what's the point of client-side validation when the user has to "click" the submit button to see the failure (same UX as server-side validation)? Seems silly. The whole benefit of client-side validation is that we can fail fast, i.e. on blur events.

We can handle it `onBlur`. Not a big deal here. Just a suggestion as an alternative approach to see if that would make life easier perhaps.

> If you think about it, errors are really just additional form state. My idea for how to handle this is that your extension component would accept two props: errors and setErrors. errors is mapping between form key and its corresponding error message (as you had suggested), and setErrors is self-explanatory. You can then use an effect to listen to the field state.

This is what I was suggesting to @jweill-aws as well. On the extension side you will have another state that handles the validity of the form itself - in the shape of an object with `[fieldName]: 'error message'` and pass the state and setter as props through the extension point. On our end we can append to this with the setter for our fields through the `onBlur` handlers for them 👍🏽

src/tokens.ts
```diff
@@ -0,0 +1,22 @@
+import { Token } from '@lumino/coreutils';
+import { ICreateJobModel, JobsView } from './model';
+
+namespace Scheduler {
+
+  export type EnvironmentParameterValue = string | number | boolean
+
+  export interface IAdvancedOptionsProps {
+    mode: JobsView
```
3coins, line 9: Are we offering an extension point for "ListJobs"?

src/tokens.ts
```diff
@@ -0,0 +1,22 @@
...
+    model: ICreateJobModel
```
3coins, line 10: This is likely going to be support in addition to `ICreateJobModel`, an analogous to the `IJobDetailModel` with all properties from the `IDescribeJob`.

3coins:
Discussed my comments offline with the team, and iterating on this in another PR is okay as this will need changes at other places as well.

3coins (approved):

PR #25, opened 2022-09-20
Material UI integration

dlqqq:
Initially authored by @ellisonbg. Per internal discussion, we would like to adopt MUI as a component framework to improve development velocity.


andrii-i:
Getting `ERROR: Could not build wheels for jupyter_scheduler, which is required to install pyproject.toml-based projects error` when trying to install package in development mode with `pip install -e .`

dlqqq:
There still exist UI regressions, but these will be handled later in separate PRs due to size and scope.

dlqqq:
Rebased off of `main`, fixed conflicts.

PR #30, opened 2022-09-21
MUI: revert breaking UI changes

dlqqq:
Fixes so far:

- fix "notebook jobs" launcher command
  - creates child widget every time parent widget needs to be re-instantiated


jweill-aws (approved):
Looks good! Thanks for your quick work on this.

dlqqq:
Had internal discussion where ownership of fixing the remaining regressions would be split evenly between team members due to size and scope.

PR #32, opened 2022-09-22
Fix create job form

jweill-aws:
WIP change to reinstate create job form functionality


3coins (approved):
Thanks for all the updates. Everything is functional now!

PR #33, opened 2022-09-22
Added compute types

3coins:
- Added optional `compute_types` in the RuntimeEnvironment
- Implemented `idempotency_token` in create_job api
- Removed default echoing of sql queries in server log output

### DIFF
--- a/jupyter_scheduler/models.py
+++ b/jupyter_scheduler/models.py
@@ -138,6 +138,7 @@ class ListJobsQuery(BaseModel):
 
 class ListJobsResponse(BaseModel):
     jobs: List[DescribeJob] = []
+    total_count: int = 0
     next_token: Optional[str] = None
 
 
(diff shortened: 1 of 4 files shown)


jupyter_scheduler/models.py
```diff
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
-from typing import Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
```
dlqqq, line 3: I think you can remove `Any`.

jupyter_scheduler/scheduler.py
```diff
@@ -135,17 +135,27 @@ def execution_manager_class(self):
...
+                p = Process(target=self.execution_manager_class(job.job_id, self.config).process)
+                p.start()
 
-            job.pid = p.pid
-            session.commit()
+                job.pid = p.pid
+                session.commit()
 
-            job_id = job.job_id
+                job_id = job.job_id
```
dlqqq, line 158: We can avoid nesting the rest of this method by returning early if `job` was already defined.

```
if job:
  return job.job_id

...
```

dlqqq (changes requested):

dlqqq (approved):

dlqqq:
@3coins Thank you! Great work.

PR #39, opened 2022-09-23
Added a script to seed fake jobs for development

3coins:
Added a python script to load fake job records for development and manual testing of list actions, for example sorting, filtering, pagination etc. Stopping these jobs or trying to open output files will not work for these jobs.
```shell
Usage: python seed.py [OPTIONS]

  Inserts random jobs in the scheduler database. Note, that this command will
  drop the tables and re-create.

Options:
  --count INTEGER  No of jobs to create, default is 25
  --db_path PATH   DB file path, default is scheduler db path
  --help           Show this message and exit.
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -20,6 +20,9 @@ prune **/node_modules
 prune lib
 prune binder
 
+# dev scripts
+prune dev
+
 # Patterns to exclude from any directory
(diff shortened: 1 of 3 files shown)


dlqqq (approved):
Looks great, thank you so much!

PR #41, opened 2022-09-23
Updated handlers to support async scheduler apis

3coins:


3coins:
Already approved in https://github.com/3coins/jupyter-scheduler/pull/1. Resubmitted here with correct base fork/branch.

PR #47, opened 2022-09-26
use MUI Table in job list view

dlqqq:
This set of changes integrates the MUI Table component into the job list view. Pagination, sorting, and loading states are correctly handled.

https://user-images.githubusercontent.com/44106031/192384710-f15af7d5-07da-430c-8c23-a355fc0b573e.mov

What this PR does NOT include:

- does not open the job details view upon clicking on job name (waiting for @andrii-i 's detail view to be merged)
- does not remove `job-detail.tsx` despite being unused, since the contents of this file may be replaced by @andrii-i

### DIFF
--- a/src/components/job-row.tsx
+++ b/src/components/job-row.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 
 import { JupyterFrontEnd } from '@jupyterlab/application';
 import { ToolbarButtonComponent } from '@jupyterlab/apputils';
@@ -15,7 +15,6 @@ import { IJobParameter, ICreateJobModel } from '../model';
 import { CommandIDs } from '..';
 
 import { replayIcon } from './icons';
-import { JobDetails } from './job-details';
 import { outputFormatsForEnvironment } from './output-format-picker';
 
 function get_file_from_path(path: string): string {
@@ -162,14 +161,6 @@ function OutputFiles(props: {
   );
 }
 
-export type JobRowProps = {
-  job: Scheduler.IDescribeJob;
-  rowClass: string;
-  cellClass: string;
-  app: JupyterFrontEnd;
-  showCreateJob: (newModel: ICreateJobModel) => void;
-};
-
 export function buildTableRow(
   job: Scheduler.IDescribeJob,
   app: JupyterFrontEnd,
@@ -179,7 +170,7 @@ export function buildTableRow(
 ): JSX.Element {
   const cellContents: React.ReactNode[] = [
     job.name,
-    job.input_uri,
+    get_file_from_path(job.input_uri),
     <OutputFiles
       job={job}
       openOnClick={(e: Event, output_uri: string) => {
@@ -187,7 +178,7 @@ export function buildTableRow(
         app.commands.execute('docmanager:open', { path: output_uri });
       }}
     />,
-    job.start_time,
+    <Timestamp job={job} />,
     translateStatus(job.status),
     <Stack spacing={1} direction="row">
       <StopButton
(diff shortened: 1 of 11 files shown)


jweill-aws:
Gut reaction from watching the video: sorting (especially by start time) changes the column widths, which I find jarring. Would it be possible to fix the column widths to avoid this?

Also, there's still an open question about whether the Action buttons should be removed (we can wait until the Job Details is in to settle this)

src/mainviews/list-jobs.tsx
```diff
@@ -52,17 +55,33 @@ export function NotebookJobsListBody(
...
+  const [page, setPage] = useState<number>(0);
+  const [maxPage, setMaxPage] = useState<number>(0);
+  const [loading, setLoading] = useState<boolean>(false);
+
+  const deleteRow = useCallback((id: Scheduler.IDescribeJob['job_id']) => {
+    setDeletedRows(deletedRows => new Set([...deletedRows, id]));
+  }, []);
+
+  const fetchInitialRows = async () => {
+    // reset pagiantion state
```
jweill-aws, line 70:
```suggestion
    // reset pagination state
```

Fix typo

dlqqq:
Rebased and resolved conflicts w/ `main`.

style/base.css
```diff
@@ -113,15 +27,7 @@ a.jp-notebook-job-name {
...
 
-.jp-notebook-jobs-list-container {
-  margin-left: 12px;
-  margin-right: 12px;
-}
-
-ul.jp-notebook-job-output-formats-options {
-  list-style: none;
-  margin: 0 0 0 -40px;
-}
```
jweill-aws, line 124: Love all this CSS class cleanup ✨

jweill-aws:
The table is sized at `0.75rem`, which seems a bit smaller than the previous version, and a little smaller than would be comfortable. What would it look like at an unmodified font size?

jweill-aws (approved):

src/mainviews/list-jobs.tsx
```diff
@@ -148,38 +191,77 @@ export function NotebookJobsListBody(
...
+              {rows.slice(page * PAGE_SIZE, (page + 1) * PAGE_SIZE)}
+            </TableBody>
+          </Table>
+          <TablePagination
+            component="div"
+            sx={{
+              position: 'sticky',
+              bottom: 0,
+              backgroundColor: 'white',
+              borderTop: '1px solid #e0e0e0'
```
3coins, line 252: These will not work with themes. We should find a way to use the JL theme vars here.

src/mainviews/list-jobs.tsx
```diff
@@ -148,38 +191,77 @@ export function NotebookJobsListBody(
...
+    .filter(job => !deletedRows.has(job.job_id))
+    .map(job =>
+      buildTableRow(
+        job,
+        props.app,
+        props.showCreateJob,
+        deleteRow,
+        translateStatus
+      )
+    );
```
3coins, line 204: Building all JSX elements here, but only using a slice of data at the bottom. Would there be a way to do the slicing of data first and only use JSX rows for the sliced data?

3coins (commented):
minor: Also noticed that the icons for sort are mis-aligned.

src/components/job-row.tsx
```diff
@@ -1,16 +1,20 @@
...
 import { closeIcon, stopIcon } from '@jupyterlab/ui-components';
 
+import Stack from '@mui/material/Stack';
+import TableRow from '@mui/material/TableRow';
+import TableCell from '@mui/material/TableCell';
+
 import { Scheduler } from '../handler';
 import { useTranslator } from '../hooks';
 import { IJobParameter, ICreateJobModel } from '../model';
+import { CommandIDs } from '..';
```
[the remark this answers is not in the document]
dlqqq: Just a shorthand for `../index` in ESM import syntax. If a source file is named `index`, only the path to its parent directory is needed.

src/mainviews/list-jobs.tsx
dlqqq: Thank you. Very observant.

src/mainviews/list-jobs.tsx
dlqqq: Again, great comment. Fixed.

PR #46, opened 2022-09-26
Add full-widget-sized job detail view

andrii-i:
This PR adds full-widget-sized job details view as per Issue #5 

Currently all fields of IDescribeJob (that extends ICreateJob) are rendered even if they are not filled. email_notifications field is not rendered

PR #48, opened 2022-09-27
Job details view

3coins:
Replacement PR for [#46](https://github.com/jupyter-server/jupyter-scheduler/pull/46)


dlqqq:
Added a commit that makes job names in the job list view clickable and take you to the corresponding detail view. 👍

PR #50, opened 2022-09-27
Added create_time, created, queued status.

3coins:
- Added create_time, update_time to CreateJob model
- Added 2 new statuses CREATED, and QUEUED. CREATED is default now when job record is created
- Corresponding updates to the UI to support create_time, CREATED status
- Removed some redundant code

**Note**: Needs a rebuild of sqlite database for this to work


dev/seed.py
```diff
@@ -18,7 +18,7 @@ def get_db_path():
 
 
 def create_random_job(index: int):
-    status = random.choice(["COMPLETED", "FAILED", "IN_PROGRESS", "STOPPED"])
+    status = random.choice(["CREATED", "COMPLETED", "FAILED", "IN_PROGRESS", "STOPPED"])
```
jweill-aws, line 21: Should `QUEUED` also be a possible choice here?

src/components/job-details.tsx
```diff
@@ -98,8 +98,8 @@ export function JobDetails(props: JobDetailsProps): JSX.Element | null {
       </div>
       <div className={'jp-notebook-job-details-grid'}>
         <div className={rowClass}>
-          <div className={keyClass}>{trans.__('Start date')}</div>
```
jweill-aws, line 101: Shouldn't both start _and_ create be displayed in the details?

jupyter_scheduler/models.py
```diff
@@ -47,6 +47,8 @@ def __str__(self) -> str:
 
 
 class Status(str, Enum):
+    CREATED = "CREATED"
+    QUEUED = "QUEUED"
```
dlqqq, line 51: When is the `QUEUED` status getting set? `CREATED` is the default status upon job creation, but I don't see when the `QUEUED` status ever gets set in the server extension. Looks like it's only defined here.

src/components/job-details.tsx
jweill-aws: Actually, this looks like the inline job details, which is disabled in the new job row — this whole file can probably be deleted if so.

src/mainviews/list-jobs.tsx
```diff
@@ -182,8 +186,8 @@ export function NotebookJobsListBody(
       name: trans.__('Output files')
     },
     {
-      sortField: 'start_time',
-      name: trans.__('Start time')
+      sortField: 'create_time',
+      name: trans.__('Create time')
```
jweill-aws, line 190: "Created" or "Created at" might be a more understandable column header for this.

dev/seed.py
3coins: Probably, we don't support on OSS side, but I can add it.

jupyter_scheduler/models.py
3coins: We don't support it, but it will be useful for other implementors who will have queueing systems. We might start using for scheduled jobs, but haven't thought about it yet.

jweill-aws (approved):

dlqqq (approved):

PR #51, opened 2022-09-27
Adds compute type to create-job form

jweill-aws:
Partial fix for #23.

After the user selects an environment, if the environment has a non-null list of compute types, display a dropdown for compute type. Serialize the compute type in the POST request to create a job.

Demo with mocked list of compute types:

https://user-images.githubusercontent.com/93281816/192654979-1557805d-a5e8-4f7a-920e-fef3a6da6ac3.mov

### DIFF
--- a/src/components/compute-type-picker.tsx
+++ b/src/components/compute-type-picker.tsx
@@ -20,7 +20,7 @@ export function ComputeTypePicker(
   const environmentObj = props.environmentList.find(
     env => env.name === props.environment
   );
-  if (!environmentObj || !environmentObj['output_formats']) {
+  if (!environmentObj || !environmentObj['compute_types']) {
     return null;
   }
 
--- a/src/components/compute-type-picker.tsx
+++ b/src/components/compute-type-picker.tsx
@@ -11,7 +11,7 @@ export type ComputeTypePickerProps = {
   environment: string;
   environmentList: Scheduler.IRuntimeEnvironment[];
   onChange: (event: SelectChangeEvent<string>) => void;
-  initialValue: string;
+  value: string | undefined;
 };
 
 export function ComputeTypePicker(
@@ -20,16 +20,13 @@ export function ComputeTypePicker(
   const environmentObj = props.environmentList.find(
     env => env.name === props.environment
   );
-  if (!environmentObj || !environmentObj['compute_types']) {
+  if (!environmentObj || !environmentObj.compute_types) {
     return null;
   }
 
-  const computeTypes = environmentObj['compute_types'] as string[];
+  const computeTypes = environmentObj.compute_types;
 
   const labelId = `${props.id}-label`;
-
-  // If no initial value was provided, default to the first value being selected.
-  const initialValue = props.initialValue || computeTypes[0];
 
   return (
     <>
@@ -39,7 +36,7 @@ export function ComputeTypePicker(
         name={props.name}
         id={props.id}
         onChange={props.onChange}
-        value={initialValue}
+        value={props.value}
       >
         {computeTypes.map((ct, idx) => (
           <MenuItem value={ct} key={idx}>
--- a/src/components/compute-type-picker.tsx
+++ b/src/components/compute-type-picker.tsx
@@ -25,7 +25,7 @@ export function ComputeTypePicker(
   }
 
   const computeTypes = environmentObj['compute_types'] as string[];
-  
+
   const labelId = `${props.id}-label`;
 
   return (
--- a/src/components/compute-type-picker.tsx
+++ b/src/components/compute-type-picker.tsx
@@ -1,0 +1,67 @@
+import React, { useMemo } from 'react';
+
+import { InputLabel, MenuItem, Select, SelectChangeEvent } from '@mui/material';
+
+import { Scheduler } from '../handler';
+
+export type ComputeTypePickerProps = {
+  label: string;
+  name: string;
+  id: string;
+  environment: string;
+  onChange: (event: SelectChangeEvent<string>) => void;
+  initialValue: string;
+};
+
+export function computeTypesForEnvironment(
+  environment: string
+): string[] | null {
+  // Retrieve the environment data from session storage.
+  const environmentsData = sessionStorage.getItem('environments');
+  if (environmentsData === null) {
+    return null;
+  }
+
+  const environments = JSON.parse(
+    environmentsData
+  ) as Array<Scheduler.IRuntimeEnvironment>;
+  const environmentObj = environments.find(env => env.name === environment);
+  if (!environmentObj || !environmentObj['output_formats']) {
+    return null;
+  }
+
+  return environmentObj['compute_types'];
+}
+
+export function ComputeTypePicker(
+  props: ComputeTypePickerProps
+): JSX.Element | null {
+  const computeTypes = useMemo(
+    () => computeTypesForEnvironment(props.environment),
+    [props.environment]
+  );
+  if (computeTypes === null) {
+    return null;
+  }
+
+  const labelId = `${props.id}-label`;
+
+  return (
+    <>
+      <InputLabel id={labelId}>{props.label}</InputLabel>
+      <Select
+        labelId={labelId}
+        name={props.name}
+        id={props.id}
+        onChange={props.onChange}
+        value={props.initialValue}
+      >
+        {computeTypes.map((ct, idx) => (
+          <MenuItem value={ct} key={idx}>
+            {ct}
+          </MenuItem>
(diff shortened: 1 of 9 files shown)


src/mainviews/create-job.tsx
```diff
@@ -288,6 +290,14 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
             environment={props.model.environment}
             value={props.model.outputFormats || []}
           />
```
dlqqq, line 292: The environments promise should not be handled within the `EnvironmentPicker` if the network request is being called from the parent. `environmentsPromise` (which really should be renamed to `fetchEnvironments` as it is a function) is being invoked every render. Look what happens when I add a console log to the function body:

https://user-images.githubusercontent.com/44106031/192658491-cc8fa074-6dff-4cd6-9a72-8cf69258d16f.mov

You need to lift this state up and make the fetch environments response run only once on initial render like so:

```
const [envList, setEnvList] = useState<Scheduler.IRuntimeEnvironment[]>([]);

...

useEffect(async () => {
  setEnvList(await environmentsPromise());
}, []);
```

Then pass `envList` to both `EnvironmentPicker` and `ComputeTypePicker`.

src/components/compute-type-picker.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return environmentObj['compute_types'];
+}
+
+export function ComputeTypePicker(
+  props: ComputeTypePickerProps
+): JSX.Element | null {
+  const computeTypes = useMemo(
+    () => computeTypesForEnvironment(props.environment),
+    [props.environment]
+  );
```
dlqqq, line 42: With the comment I made earlier, `ComputeTypePicker` can now expect a deserialized `Scheduler.IRuntimeEnvironment[]` directly as an additional prop (I suggest `envList`) Then, you can just inline the body of `computeTypesForEnvironment` within `useMemo` since it's fairly simple, and delete its original definition.

```
const computeTypes = useMemo(() => {
  const targetEnv = props.envList.find(env => env.name === props.environment);
  return targetEnv?.output_formats ? targetEnv.compute_types : null;
}, [props.envList, props.environment])
```

I would also rename the `environment` prop to `envName` to further emphasize its type as a string and to distinguish it from `envList`.

dlqqq (changes requested):

src/mainviews/create-job.tsx
jweill-aws: Thank you! Piyush was just asking about this.

jweill-aws: The `useEffect` function can't take an `async` function — I get an error when I try to use the code you provided. There is a workaround, but be warned: it's ugly even by JavaScript standards. https://stackoverflow.com/questions/56838392/how-to-call-an-async-function-inside-a-useeffect-in-react

src/components/compute-type-picker.tsx
jweill-aws: Is `useMemo` still necessary? Now that we expect the props to be passed in and populated from the get-go, I had success removing it.

src/handler.ts
```diff
@@ -338,6 +339,7 @@ export namespace Scheduler {
     file_extensions: string[];
     output_formats: IOutputFormat[];
     metadata: { [key: string]: any };
```
3coins, line 341:
```suggestion
    metadata: { [key: string]: string };
```

src/model.ts
```diff
@@ -84,6 +84,7 @@ export interface ICreateJobModel {
   environment: string;
   parameters?: IJobParameter[];
   outputFormats?: IOutputFormat[];
+  computeType?: string;
```
3coins, line 87: How about the rest of the model properties, for example `idempotency_token`, `tags` etc.?

3coins:
@jweill-aws 
Thanks for fixing the environment load issue.
Should we default to first value for compute type? Currently the picker shows blank, so users have to select a value, but you can still submit without selecting one, which will be invalid.

3coins (changes requested):

src/model.ts
jweill-aws: I was planning to add those when I added the UI controls/display for them.

src/handler.ts
jweill-aws: Could any metadata item be, for example, an array?

3coins: No, the intent of metadata was to provide simple values for display.

src/mainviews/create-job.tsx
dlqqq: Oh yeah, forgot about that quirk of useEffect callbacks.

src/components/compute-type-picker.tsx
dlqqq: useMemo shouldn't ever be "necessary" unless you are doing something evil like depending on only evaluating an expression on certain renders but not others. It's an optimization that basically only re-evaluates its callback whenever one of the dependencies in the second argument changes from the previous render (including the initial render). I personally prefer to use it not just as an optimization but also as a way of isolating intermediate variables (like `targetEnv`) into its own block scope.

src/components/compute-type-picker.tsx
```diff
@@ -9,43 +9,28 @@ export type ComputeTypePickerProps = {
...
     return null;
   }
 
+  const computeTypes = environmentObj['compute_types'] as string[];
+
   const labelId = `${props.id}-label`;
 
+  // If no initial value was provided, default to the first value being selected.
+  const initialValue = props.initialValue || computeTypes[0];
+
```
dlqqq, line 33: I just realized how poorly named `props.initialValue` is. It's not an initial value, it actually does set the value for this field. So, there's a few things wrong here.

1) Rename `props.initialValue` to `props.value`.

2) The default value of a field should never be set by the field itself, since there's no way for the parent form component to read it and actually submit it to the backend. This is essentially just for rendering purposes only and is actually quite misleading, since `props.model.computeType` (the value actually sent to the backend) can still be nullish.

Thus, the default value needs to be lifted up and immediately set by the form after `environment` gets selected. In `create-job.tsx`:

```
  const handleSelectChange = (event: SelectChangeEvent<string>) => {
    const target = event.target as HTMLInputElement;

    // if setting environment, default to first compute type
    if (target.name === 'environment') {
      const envObj = environmentList.find(env => env.name === target.value)
      props.handleModelChange({ ...props.model, environment: target.value, computeType: envObj.compute_types?.[0] })
    else {
      // otherwise just set the model
      props.handleModelChange({ ...props.model, [target.name]: target.value });
    }
  };
```

src/mainviews/create-job.tsx
```diff
@@ -295,6 +298,7 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
             name="computeType"
             id={`${formPrefix}computeType`}
             onChange={handleSelectChange}
+            environmentList={environmentList}
             environment={props.model.environment}
             initialValue={props.model.computeType || ''}
```
dlqqq, line 303: Remove the default value of `''` here since we will be setting the default as soon as environment gets selected.

src/mainviews/create-job.tsx
```diff
@@ -58,10 +58,26 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+    Scheduler.IRuntimeEnvironment[]
+  >([]);
+
   // A mapping from input names to error messages.
   // If an error message is "truthy" (i.e., not null or ''), we should display the
   // input in an error state and block form submission.
-  const [errors, setErrors] = React.useState<SchedulerTokens.ErrorsType>({});
+  const [errors, setErrors] = useState<SchedulerTokens.ErrorsType>({});
+
+  const api = new SchedulerService({});
```
dlqqq, line 71: This creates a new `SchedulerService` every render, which is especially bad for forms where a render is triggered every time a key is pressed.

Right now, the SS constructor isn't doing much, so this doesn't have too much of an impact yet. But let's not have our form performance have an implicit dependency on the SS constructor.

```
const api = useMemo(() => new SchedulerService({}), []);
```

dlqqq (changes requested):

src/components/compute-type-picker.tsx
dlqqq: Also, keep in mind that this control flow is necessary because you should only be invoking `handleModelChange` once per render. This is because we're not use the callback syntax of `setState`, meaning that:

```
props.handleModelChange({ ...props.model, a: "asdf" })
props.handleModelChange({ ...props.model, b: "asdf" })
```

actually only sets property `b`. This is because within the lifetime of this function, `props.model` refers to the same object.

PR #56, opened 2022-09-28
Adds idempotency token and tags for use in create job, details view

jweill-aws:
Fixes #23.

Also modifies the details view to incorporate the extension point.

Create job form:

image

Details page:

image

PR #60, opened 2022-09-29
Job Detail refactor

andrii-i:
This PR refactors Job Detail, migrates styling to two-column flex-based adaptable layout, uses Cards to section page content, adds links to output files

Preview:
job-detail_flex_links

### DIFF
--- a/src/mainviews/job-detail.tsx
+++ b/src/mainviews/job-detail.tsx
@@ -259,8 +259,8 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
         </FormLabel>
         <Stack spacing={4}>
           {props.model.parameters &&
-            props.model.parameters.map(parameter => (
-              <Stack direction={'row'} gap={2} flexWrap={'wrap'}>
+            props.model.parameters.map((parameter, idx) => (
+              <Stack key={idx} direction={'row'} gap={2} flexWrap={'wrap'}>
                 <TextFieldStyled
                   label={trans.__('Parameter name')}
                   defaultValue={parameter.name}
--- a/src/mainviews/job-detail.tsx
+++ b/src/mainviews/job-detail.tsx
@@ -1,15 +1,19 @@
 import React, { useEffect, useState } from 'react';
+
+import { JupyterFrontEnd } from '@jupyterlab/application';
 import {
   convertDescribeJobtoJobDetail,
   ICreateJobModel,
   IJobDetailModel,
   JobsView
 } from '../model';
+
 import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
 import Breadcrumbs from '@mui/material/Breadcrumbs';
 import Link from '@mui/material/Link';
 import Stack from '@mui/material/Stack';
-import Button from '@mui/material/Button';
+
 import {
   Card,
   CardContent,
@@ -22,7 +26,6 @@ import {
 import { useTranslator } from '../hooks';
 import { Heading } from '../components/heading';
 import { SchedulerService } from '../handler';
-import { JupyterFrontEnd } from '@jupyterlab/application';
 import { Scheduler as SchedulerTokens } from '../tokens';
 
 export interface IJobDetailProps {
--- a/src/mainviews/job-detail.tsx
+++ b/src/mainviews/job-detail.tsx
@@ -112,7 +112,7 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
   );
 
   const ButtonBar = () => (
-    <Stack direction="row" spacing={1} justifyContent="flex-end">
+    <Stack direction="row" gap={2} justifyContent="flex-end" flexWrap={'wrap'}>
       {job?.status === 'IN_PROGRESS' && (
         <Button variant="outlined" onClick={handleStopJob}>
           {trans.__('Stop Job')}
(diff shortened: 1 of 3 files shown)


src/mainviews/job-detail.tsx
```diff
@@ -109,203 +94,172 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+  const Parameters = () => (
+    <Card>
+      <CardContent>
+        <FormLabel sx={{ mb: 4 }} component="legend">
+          {trans.__('Parameters')}
+        </FormLabel>
+        <Grid container spacing={4}>
+          {Object.entries(job?.parameters ?? {}).map(([parameter, value]) => (
+            <React.Fragment key={parameter}>
+              <Grid xs={12} md={6}>
```
jweill-aws, line 176: The docs recommend using `<Grid item` inside a `<Grid container`.

src/mainviews/job-detail.tsx
```diff
@@ -109,203 +94,172 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
                 <TextFieldStyled
-                  label={trans.__('Start time')}
-                  defaultValue={timestampLocalize(job?.start_time ?? '')}
+                  label={trans.__('Parameter')}
```
jweill-aws, line 178: `Parameter name` seems more appropriate here

src/mainviews/job-detail.tsx
```diff
@@ -109,203 +94,172 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+                  defaultValue={parameter}
                 />
+              </Grid>
+              <Grid xs={12} md={6}>
                 <TextFieldStyled
-                  label={trans.__('End time')}
-                  defaultValue={timestampLocalize(job?.end_time ?? '')}
+                  label={trans.__('Value')}
```
jweill-aws, line 184: `Parameter value`?

src/mainviews/job-detail.tsx
```diff
@@ -109,203 +94,172 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
-          <Heading level={1}>{trans.__('Job Detail')}</Heading>
-          <MainArea />
+          <TextFieldStyled
+            label={trans.__('Idempotency token')}
+            defaultValue={job?.idempotency_token ?? ''}
+          />
+          <FormLabel component="legend">{trans.__('Tags')}</FormLabel>
+          {job?.tags &&
+            job?.tags.map(tag => (
+              <TextFieldStyled label={trans.__('Tag')} defaultValue={tag} />
```
jweill-aws, line 206: See #56 — the label `Tag` for every tag is redundant. Perhaps Iterate on `(tag, idx)` and use `idx + 1` (index plus one) in the tag name: `trans.__('Tag %1', idx + 1)`. This will make each successive tag have a distinct title.

src/mainviews/job-detail.tsx
andrii-i: Agree. Changed

src/mainviews/job-detail.tsx
andrii-i: It is correct for [Grid](https://mui.com/material-ui/react-grid/). [Grid 2 ](https://mui.com/material-ui/react-grid2/)(or `@mui/material/Unstable_Grid2`) gets rid of item prop and some other deficiencies. I decided on using grid based on the fact that it gives clearer code and and that mui website lists it as 'new implementation'. I do agree that it might be better to use stable Grid version for the sake of risk aversion:
Screen Shot 2022-09-29 at 9 52 25 AM

src/mainviews/job-detail.tsx
andrii-i: That's a good suggestion for readability, implemented

src/mainviews/job-detail.tsx
```diff
@@ -106,206 +104,190 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+        </Link>
+        <Typography color="text.primary">
+          {job?.name ?? props.model.jobId}
+        </Typography>
+      </Breadcrumbs>
+    </div>
+  );
+
+  return (
+    <Box sx={{ p: 4 }} style={{ height: '100%', boxSizing: 'border-box' }}>
```
dlqqq, line 275: Do not set the `height` property here. This is causing the bottom padding to not be applied to the list view body.

dlqqq (changes requested):

src/mainviews/job-detail.tsx
dlqqq: Fixed in latest commit.

andrii-i:
Refactor completed, ready for review. Please see latest preview in the PR description

src/mainviews/job-detail.tsx
andrii-i: Good suggestion, thank you

src/mainviews/job-detail.tsx
```diff
@@ -120,201 +96,235 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+      {
+        defaultValue: props.model.environment,
+        label: trans.__('Environment')
+      },
+      { defaultValue: props.model.status ?? '', label: trans.__('Status') }
+    ],
+    [
+      {
+        defaultValue: timestampLocalize(props.model.createTime ?? ''),
+        label: trans.__('Create time')
```
jweill-aws, line 169:
```suggestion
        label: trans.__('Created at')
```

Consistency with the jobs list's column header

src/mainviews/job-detail.tsx
```diff
@@ -120,201 +96,235 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+      },
+      {
+        defaultValue: timestampLocalize(props.model.updateTime ?? ''),
+        label: trans.__('Update time')
```
jweill-aws, line 173:
```suggestion
        label: trans.__('Updated at')
```

Consistency with `Created at` above

src/mainviews/job-detail.tsx
```diff
@@ -120,201 +96,235 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+  const Parameters = () => (
+    <Card>
+      <CardContent>
+        <FormLabel sx={{ mb: 4 }} component="legend">
+          {trans.__('Parameters')}
+        </FormLabel>
+        <Stack spacing={4}>
+          {props.model.parameters &&
+            props.model.parameters.map(parameter => (
+              <Stack direction={'row'} gap={2} flexWrap={'wrap'}>
```
jweill-aws, line 263: There should be a unique `key` for each component created this way. You can do this by adding the index (`idx`) to the `map` function:

```suggestion
            props.model.parameters.map((parameter, idx) => (
              <Stack key={idx} direction={'row'} gap={2} flexWrap={'wrap'}>
```

andrii-i: Thank you!

src/mainviews/job-detail.tsx
```diff
@@ -120,201 +96,235 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+          e.preventDefault();
+          props.app.commands.execute('docmanager:open', {
+            path: outputName
+          });
+        }}
+        style={{ paddingRight: '1em' }}
+      >
+        {props.outputType}
+      </Link>
     );
```
3coins, line 214: This looks like a duplication of https://github.com/jupyter-server/jupyter-scheduler/blob/9623dceb73889fd44689208d7bcb32888521e854/src/components/job-row.tsx#L133-L164. Can you re-use this?

andrii-i: It's essentially the same function adapted to the Job Detail view needs / types used in the view. Original function in `job-row` accepts `Scheduler.IDescribeJob`, type that API call returns while adapted function (in job-detail) accepts `IJobDetailModel`, same type it's getting passed to and maps all its fields from. I do call API from inside the function so as an option I could save API return object in the state but it would somewhat duplicate the state and go against the principle that this component works with "front-end types"

dlqqq (changes requested):

src/mainviews/job-detail.tsx
```diff
@@ -120,201 +96,235 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
     }
   };
 
-  useEffect(() => {
-    getJob();
-  }, []);
+  const TextFieldStyled = (props: TextFieldProps) => (
+    <TextField {...props} variant="outlined" disabled />
+  );
```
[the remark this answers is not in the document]
andrii-i: As discussed, moved out of main function / JobDetail scope

src/mainviews/job-detail.tsx
3coins: Job Detail doesn't have anything additional that list doesn't have.  Which API call are you referring to?

src/mainviews/job-detail.tsx
```diff
@@ -120,201 +96,235 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
...
+        </Button>
+      )}
+      <Button variant="outlined" onClick={handleRerunJob}>
+        {trans.__('Rerun Job')}
+      </Button>
+      <Button variant="contained" color="error" onClick={handleDeleteJob}>
+        {trans.__('Delete Job')}
+      </Button>
+    </Stack>
+  );
```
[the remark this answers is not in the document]
andrii-i: Defined all elements within the component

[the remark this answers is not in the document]
andrii-i: But it doesn't work when using in return. E.g. when `{Loading}` in element's return()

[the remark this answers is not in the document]
andrii-i: Tried implementing this but this way of implementing `JSX.Framgment` as a `const` doesn't seem to work inside the component `return()`

Ways I tried to make it work in `return()` (all error out or render white screen):
image
image
image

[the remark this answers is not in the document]
andrii-i: Changed all logical sections of the element to be elements within the component v sub-components how it was before

src/mainviews/job-detail.tsx
andrii-i: To the call currently at line 56. Created [issue 56](https://github.com/jupyter-server/jupyter-scheduler/issues/75) to track it

dlqqq (approved):

PR #62, opened 2022-09-29
Additional padding to cause icon buttons to be circular

jweill-aws:
Add and delete buttons are now circular, not 🥚 -shaped.

image

image

### DIFF
--- a/src/components/icon-buttons.tsx
+++ b/src/components/icon-buttons.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+
+import { addIcon, closeIcon } from '@jupyterlab/ui-components';
+import { IconButton } from '@mui/material';
+
+export interface IDeleteButtonProps {
+  onClick: React.MouseEventHandler<HTMLButtonElement> | undefined;
+  title: string;
+}
+
+// Exaggerate horizontal padding to make the button circular, not oblate
+const extraHorizontalPadding = { px: '12.5px', py: '8px' };
+
+export function DeleteButton(props: IDeleteButtonProps): JSX.Element {
+  return (
+    <IconButton
+      aria-label="delete"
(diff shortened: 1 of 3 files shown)


src/components/icon-buttons.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { addIcon, closeIcon } from '@jupyterlab/ui-components';
+import { IconButton } from '@mui/material';
+
+export interface IDeleteButtonProps {
+  onClick: React.MouseEventHandler<HTMLButtonElement> | undefined;
+  title: string;
+}
+
+// Exaggerate horizontal padding to make the button circular, not oblate
+const extraHorizontalPadding = { px: '12.5px', py: '8px' };
```
dlqqq, line 12: The problem now is that the icon button is just way too big relative to the size of the icon itself. The reason is that you're not actually fixing the root issue of why the icon button is egg shaped. `svg` elements have `display: inline` set by default. Hence the `div` wrapping the `svg` has a minimum height set by the line height being specified by CSS. Check this demo out.


https://user-images.githubusercontent.com/44106031/193149185-0bf54dce-9c37-4f5e-9ecf-b7699528a1ea.mov


There are two options here:

1. Set `line-height: 0` in the parent element `<IconButton />`
2. Set `display: block` on the SVG icons.

And of course, remove the extra horizontal padding.

dlqqq (changes requested):

src/components/icon-buttons.tsx
jweill-aws: I'll miss the novelty oversized buttons, but `line-height: 0` worked well for me. Thanks!

dlqqq (approved):
Looks great!

PR #63, opened 2022-09-29
call renderDOM() synchronously in NotebookJobsPanel

### LINKED ISSUE #34: Editing parameters causes cursor to jump to end of input
## Description

While editing the name or value of a parameter, the cursor jumps to the end of the input box after every character entry. This is present in the code in #32.

https://user-images.githubusercontent.com/93281816/191857029-51e5d3c2-a5ff-4394-9504-49dc62870dbf.mov

## Reproduce

Click in the middle of the name or value of a parameter. Type a character. The character is inserted at that position and the cursor moves to the end of the input.

## Expected behavior

The cursor appears immediately following the inserted character.

## Context

This does not occur with other text inputs in this form.

This appears to be common behavior with React when the state update is not fast enough. I attempted to remediate this using local state, as mentioned below, but I didn't have the same success with the parameter inputs as I did with the standalone inputs.

* https://github.com/facebook/react/issues/955
* https://stackoverflow.com/questions/28922275/in-reactjs-why-does-setstate-behave-differently-when-called-synchronously/28922465#28922465

dlqqq:
Fixes #34.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -132,6 +132,12 @@ export class JobsModel extends VDomModel {
   private _createJobModel: ICreateJobModel;
   private _listJobsModel: IListJobsModel;
   private _jobDetailModel: IJobDetailModel;
+  /**
+   * Callback that gets invoked whenever a model is updated. This should be used
+   * to call `ReactWidget.renderDOM()` to synchronously update the VDOM rather
+   * than triggering an async VDOM update via Lumino. See #34.
+   */
+  private _onModelUpdate?: () => unknown;
   private _jobCount: number;
 
   constructor(options: IJobsModelOptions) {
@@ -143,6 +149,7 @@ export class JobsModel extends VDomModel {
       ...Private.emptyCreateModel(),
       jobId: ''
     };
+    this._onModelUpdate = options.onModelUpdate;
     this._jobCount = 0;
   }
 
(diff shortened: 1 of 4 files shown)


src/model.ts
```diff
@@ -132,6 +132,12 @@ export class JobsModel extends VDomModel {
   private _createJobModel: ICreateJobModel;
   private _listJobsModel: IListJobsModel;
   private _jobDetailModel: IJobDetailModel;
+  /**
+   * Callback that gets invoked whenever a model is updated. This should be used
+   * to call `ReactWidget.renderDOM()` to synchronously update the VDOM rather
+   * than triggering an async VDOM update via Lumino. See #34.
```
jweill-aws, line 138:
```suggestion
   * than triggering an async VDOM update via Lumino.
```

Not necessary to refer to issues or PRs in code comments

dlqqq: Why not? Issue IDs are immutable and can add a lot of valuable context that can't fit inside of a code comment.

dlqqq: I left a comment on that issue that describes in detail why this issue occurs and why this workaround is necessary. That level of detail is valuable for future devs.

jweill-aws (approved):

src/model.ts
jweill-aws: The problem is that "See #34" outside the context of a GitHub issue or PR just means "see number thirty-four." Especially as code gets copied and reused, this will mean less over time.

You could provide a more obvious cue with a URL, although I've never found it popular to include URLs in source code.

dlqqq: I think this comment is sufficiently obvious to developers. Using my discretion here and merging.

PR #65, opened 2022-09-30
Adds validation: parameter name must be specified if value is specified

jweill-aws:
In the create form, if the parameter value is nonblank, the parameter name must not be blank.

If the parameter name and value are both blank, there is no error.

If the parameter name is nonblank and the value is blank, there is no error.

Errors are displayed on name or value blur.

There is no validation for the value.

https://user-images.githubusercontent.com/93281816/193168959-4eb43689-e8bf-447f-8f91-3098c5e45566.mov

### DIFF
--- a/src/components/parameters-picker.tsx
+++ b/src/components/parameters-picker.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, useCallback, useState } from 'react';
+import React, { ChangeEvent } from 'react';
 
 import { InputLabel, TextField } from '@mui/material';
 import Stack from '@mui/system/Stack';
@@ -23,18 +23,8 @@ export type ParametersPickerProps = {
   handleErrorsChange: (errors: Scheduler.ErrorsType) => void;
 };
 
-const useForceUpdate = () => {
-  const [, setTick] = useState(0);
-  const update = useCallback(() => {
-    setTick(tick => tick + 1);
-  }, []);
-  return update;
-};
-
 export function ParametersPicker(props: ParametersPickerProps): JSX.Element {
   const trans = useTranslator('jupyterlab');
-
-  const update = useForceUpdate();
 
   const checkParameter = (
     e: EventTarget & (HTMLInputElement | HTMLTextAreaElement)
@@ -42,29 +32,20 @@ export function ParametersPicker(props: ParametersPickerProps): JSX.Element {
     const paramInputName = e.name;
 
     const paramMatch = paramInputName.match(/^parameter-(\d+)/);
-    if (paramMatch === null) {
+    if (!paramMatch || paramMatch.length < 2) {
       return; // Invalid parameter name; should not happen
     }
     const paramIdx = parseInt(paramMatch[1]);
 
-    const modifiedParam = props.value[paramIdx];
-    const paramName = modifiedParam.name;
-    const paramValue = modifiedParam.value;
+    const param = props.value[paramIdx];
+    const invalid = param.name === '' && param.value !== '';
 
-    if (paramName === '' && paramValue !== '') {
-      const newErrors: Scheduler.ErrorsType = props.errors;
-      newErrors['parameter-' + paramIdx + '-name'] = trans.__(
-        'No name specified for this parameter.'
-      );
-      props.handleErrorsChange(newErrors);
-    } else {
-      const newErrors: Scheduler.ErrorsType = props.errors;
-      newErrors['parameter-' + paramIdx + '-name'] = '';
-      props.handleErrorsChange(newErrors);
-    }
-
-    // Force a rerender.
-    update();
+    props.handleErrorsChange({
+      ...props.errors,
+      [`parameter-${paramIdx}-name`]: invalid
+        ? trans.__('No name specified for this parameter.')
+        : ''
+    });
   };
 
   return (
--- a/src/components/parameters-picker.tsx
+++ b/src/components/parameters-picker.tsx
@@ -77,11 +77,10 @@ export function ParametersPicker(props: ParametersPickerProps): JSX.Element {
               onBlur={e => checkParameter(e.target)}
               onChange={props.onChange}
             />
-            {/* The addedStyle will probably have to be changed again once MUI is restyled to have smaller text box height */}
             <DeleteButton
               onClick={() => props.removeParameter(paramIdx)}
               title={trans.__('Delete this parameter')}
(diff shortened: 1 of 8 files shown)


3coins:
@jweill-aws 
I am not sure about allowing a parameter name with blank value. It would seem that in most cases this is likely an oversight by the user and they probably missed providing a value; prompting user to add a value in this case is likely the right intent. What do you think?

src/components/parameters-picker.tsx
```diff
@@ -18,39 +19,95 @@ export type ParametersPickerProps = {
...
+  handleErrorsChange: (errors: Scheduler.ErrorsType) => void;
+};
+
+const useForceUpdate = () => {
+  const [, setTick] = useState(0);
+  const update = useCallback(() => {
+    setTick(tick => tick + 1);
+  }, []);
+  return update;
 };
```
3coins, line 32: Why is this required?

jweill-aws: Without `useForceUpdate`, although the model is updated, the parameters component does not automatically rerender. This function is meant to update a nominal state to force a rerender.

jweill-aws:
> @jweill-aws I am not sure about allowing a parameter name with blank value. It would seem that in most cases this is likely an oversight by the user and they probably missed providing a value; prompting user to add a value in this case is likely the right intent. What do you think?

A parameter with a name and an empty-string value is not inherently an error; there might be a reasonable reason for that. (The `errors` variable uses an empty-string value to mean "no error" for a particular input name, for example.) @ellisonbg had suggested the specific case of "empty parameter name, non-empty parameter value" being an error, but I don't consider the other three cases to be problematic:

| | Empty name | Name |
|-|-|-|
| **Empty value** | ✅  | ✅  |
| **Value** | ❌  | ✅  |

src/components/parameters-picker.tsx
3coins: Do we know why this behavior is happening? 
That seems like a hack to support a refresh that React supports by design; this is likely an issue with our code. If `ParametersPicker` is a functional component, any update to props is supposed to re-render it.  If you need internal state in the `ParametersPicker` to trigger changes, then that's the way to go. In either case, there shouldn't be a pattern of adding an unused state to force updates.

3coins: > A parameter with a name and an empty-string value is not inherently an error; there might be a reasonable reason for that. (The `errors` variable uses an empty-string value to mean "no error" for a particular input name, for example.) @ellisonbg had suggested the specific case of "empty parameter name, non-empty parameter value" being an error, but I don't consider the other three cases to be problematic:
> 
> Empty name	Name
> **Empty value**	✅	✅
> **Value**	❌	✅

From a user's perspective, there is a very low chance they would want to pass an empty string for a parameter; for other cases (number, boolean) this will cause an exception. On the other hand, blocking users to pass an empty value will hinder some users, I don't feel that we will have a lot of users who would want this. Ok to submit this as-is and revisit after some user feedback.

dlqqq: BTW, the async update was fixed in #63.

dlqqq:
> A parameter with a name and an empty-string value is not inherently an error; there might be a reasonable reason for that.

I believe this is to preserve parity with shell environment variables, which can be defined without values. A common example is just to define `export DEBUG` in your local shell and have an application dispatch additional logging based on that.

src/components/parameters-picker.tsx
```diff
@@ -18,39 +19,95 @@ export type ParametersPickerProps = {
...
+      );
+      props.handleErrorsChange(newErrors);
+    } else {
+      const newErrors: Scheduler.ErrorsType = props.errors;
+      newErrors['parameter-' + paramIdx + '-name'] = '';
+      props.handleErrorsChange(newErrors);
+    }
+
+    // Force a rerender.
+    update();
```
dlqqq, line 67: The reason you needed to force a re-render here is that React detects state changes based on object identity. Here you're taking `props.errors` and adding/modifying a property, rather than creating a new `errors` object.

dlqqq (changes requested):

src/components/parameters-picker.tsx
jweill-aws: Shouldn't `handleErrorsChange`, aka `setErrors`, be modifying the state based on the modified/added property?

dlqqq:
Other bugs I found while playing around:

https://user-images.githubusercontent.com/44106031/193327050-06e5314a-eaee-4144-83b8-57f6d2ce7422.mov

Deleting param should clear the corresponding errors.

https://user-images.githubusercontent.com/44106031/193327448-7a142d7a-c95a-411b-a765-f3e77ffa9d73.mov

Error state is not being correctly re-calculated upon deleting a param in the middle.

dlqqq:
@jweill-aws @3coins I rebased off of `main`, removed the `useForceUpdate()` hook, and fixed the aforementioned error state bugs.

src/mainviews/create-job.tsx
```diff
@@ -216,7 +216,31 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+      }
+
+      // otherwise, restore errors with params after deleted param by offsetting
+      // their index by -1
+      newErrors[`parameter-${paramIdx - 1}-name`] =
+        errors[`parameter-${paramIdx}-name`];
+    }
+
     props.handleModelChange({ ...props.model, parameters: newParams });
+    setErrors(newErrors);
```
dlqqq, line 243: You can see how messy maintaining parameter error state consistency can get. Here's my suggestion to you. Have you considered making the parameters error state a nested property of `errors`? For example, `errors.parameters` would be an array of error strings for each parameter. Then, the logic here can be simplified to:

```ts
const newParams = props.model.parameters || [];
newParams.splice(idx, 1);

const newParamErrors = errors.parameters || [];
newParamErrors.splice(idx, 1);

props.handleModelChange({ ...props.model, parameters: newParams });
setErrors((errors) => ({ ...errors, parameters: newParamErrors }));
```

dlqqq (changes requested):

src/components/parameters-picker.tsx
dlqqq: Nope, it seems to use `Object.is()` to determine whether to trigger an update to the VDOM. Now, I tried finding documentation for this and I honestly can't find anything official; just a bunch of SO answers. Consider this secret FEE lore. 😅

dlqqq (approved):

Issue #79, opened 2022-10-04
runtime_environment_parameters not passed to API endpoint on form submission


jweill-aws:
## Description

The value of `runtime_environment_parameters` is not passed to the API endpoint on form submission.

## Reproduce

Create a job using the advanced options extension point that modifies `runtime_environment_parameters` in the model.

## Expected behavior

`runtime_environment_parameters` is present in the API call.

jweill-aws:
Fixed by #82 and #83.

jweill-aws:
Fix released in 0.3.2.

PR #82, opened 2022-10-04
Serializes runtime_environment_parameters in model, request

jweill-aws:
Fix for #79 as requested by @awaisabir. Adds `runtime_environment_parameters` to request and `runtimeEnvironmentParameters` to model.


awaisabir (approved):
I believe making this available in the model should be good enough. Thanks 👍🏽

PR #100, opened 2022-10-06
Hides idempotency token for JobSchedule

### LINKED ISSUE #99: Idempotency token should not be shown when user chooses "Run on a schedule"
## Description

The idempotency token is not part of the schema for a job definition, but the "Idempotency token" continues to appear in the create form after the user switches modes to "Run on a schedule"

image

## Expected behavior

After the user switches to "Run on a schedule," no longer show the idempotency token input. When the user switches back to "Run now," show the idempotency token input.

jweill-aws:
Fixes #99.

For job creation, continue to show idempotency token:

image

For job definition creation, do not show idempotency token:

image

### DIFF
--- a/src/advanced-options.tsx
+++ b/src/advanced-options.tsx
@@ -119,17 +119,20 @@ const AdvancedOptions = (
   const tagsDisplay: JSX.Element | null =
     props.jobsView === 'CreateJob' ? createTags() : showTags();
 
+  // The idempotency token is only used for jobs, not for job definitions
   return (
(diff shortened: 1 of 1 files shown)


dlqqq:
Looks good. UI also doesn't send the idempotency token to the backend even if you fill it in before switching to "run on a schedule", thanks to the logic in `submitCreateJobDefinitionRequest()`.

dlqqq (approved):

PR #102, opened 2022-10-06
Add job definitions to DB seed script

dlqqq:
Also fixes bug encountered where jobs cannot be filtered by job definition ID.

### DIFF
--- a/dev/seed.py
+++ b/dev/seed.py
@@ -2,10 +2,12 @@ import json
 import os
 import random
 import subprocess
+import zoneinfo
+from uuid import uuid4
 
 import click
 
-from jupyter_scheduler.orm import Job, create_session, create_tables
+from jupyter_scheduler.orm import Job, JobDefinition, create_session, create_tables
 from jupyter_scheduler.utils import get_utc_timestamp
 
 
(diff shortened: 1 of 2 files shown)


dev/seed.py
```diff
@@ -17,7 +19,39 @@ def get_db_path():
...
+            "lame definition",
+        ]
+    )
+    job_definition_id = str(uuid4())
+    input_uri = "".join(name.split()) + ".ipynb"
+    output_prefix = ""
+    timezone = random.choice(list(zoneinfo.available_timezones()))
+    # random schedules can be generated via https://crontab.guru/
+    schedule = random.choice(
+        ["0 0,12 1 */2 *", "0 4 8-14 * *", "0 0 1,15 * 3", "@weekly", "5 0 * 8 *", "15 14 1 * *"]
```
3coins, line 38: Is "@weekly" a failure case?

dev/seed.py
```diff
@@ -67,29 +102,43 @@ def load_data(count: int, db_path: str):
...
+@click.option("--jobs-count", default=57, help="No of jobs to create, default is 57.")
+@click.option(
+    "--job-defs-count", default=27, help="No of job definitions to create, default is 27."
+)
 @click.option(
     "--db_path",
     type=click.Path(),
     default=get_db_path(),
     help="DB file path, default is scheduler db path",
 )
```
3coins, line 139: Can you check if there are shortcut options (e.g., -j -jd -db) for inputs and add them?

jupyter_scheduler/scheduler.py
```diff
@@ -191,7 +191,7 @@ def list_jobs(self, query: ListJobsQuery) -> ListJobsResponse:
             if query.status:
                 jobs = jobs.filter(Job.status == query.status)
             if query.job_definition_id:
-                jobs = jobs.filter(Job.status == query.job_definition_id)
+                jobs = jobs.filter(Job.job_definition_id == query.job_definition_id)
```
3coins, line 194: Thanks for this.

3coins (commented):
Looks good. Some minor comments.

dev/seed.py
dlqqq: It was generated by crontab.guru, but it looks like it's non-standard. I'll remove it from the schedule stubs.

dev/seed.py
dlqqq: Doing this is easy just by adding an additional positional argument to `@click.option()`. One small change though, we will be using double-hyphen (`--j`, `--jd`, `--db`) to specify these CLI arguments require an additional CLI argument after them, to conform to existing conventions.

https://serverfault.com/questions/387935/whats-the-difference-betwen-the-single-dash-and-double-dash-flags-on-shell-comm

3coins (approved):

Issue #98, opened 2022-10-06
Display job definition schedule in a human-readable way


jweill-aws:
### Problem

The job definition schedule is stored in cron format (e.g., `*/5 * * * *`) which most users will not recognize in text.

### Proposed Solution

In the job definition list and details view, display the job definition schedule in a human-readable way (e.g., `*/5 * * * *` as "every 5 minutes").

### Additional context

All supported JupyterLab languages should be supported; the user's chosen locale should be used for display.

cron explanation libraries exist for multiple languages.

* Python: https://github.com/Salamek/cron-descriptor
* JavaScript: https://github.com/bradymholt/cRonstrue

PR #103, opened 2022-10-06
Shows helper text for cron expression

jweill-aws:
Partial fix for #98.

Displays the cron expression helpfully for the user. Adds a few hint buttons to get the user started with cron syntax.

https://user-images.githubusercontent.com/93281816/194433098-6fa50458-ea8f-42a4-93f6-7407dbd69d38.mov

The text box's helper text updates dynamically with interpretations of valid cron expressions:

https://user-images.githubusercontent.com/93281816/194433134-0052a870-79d5-435e-a53a-90a6f4a39c4c.mov

This is not yet localized; it always displays in English, although the library chosen supports other languages.

### DIFF
--- a/src/components/schedule-inputs.tsx
+++ b/src/components/schedule-inputs.tsx
@@ -1,4 +1,6 @@
 import React, { ChangeEvent } from 'react';
+
+import cronstrue from 'cronstrue';
 import tzdata from 'tzdata';
 
 import { Autocomplete, TextField } from '@mui/material';
@@ -23,6 +25,16 @@ export function ScheduleInputs(props: ScheduleInputsProps): JSX.Element | null {
 
   const timezoneLabel = trans.__('Time zone');
 
+  let cronString;
+  try {
+    if (props.schedule !== undefined && !props.errors['schedule']) {
+      cronString = cronstrue.toString(props.schedule);
+    }
+  }
+  catch (e) {
+    // Do nothing; let the errors or nothing display instead
+  }
+
   return (
     <>
       <TextField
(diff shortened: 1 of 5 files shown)


src/components/schedule-inputs.tsx
```diff
@@ -23,6 +25,16 @@ export function ScheduleInputs(props: ScheduleInputsProps): JSX.Element | null {
...
   const timezoneLabel = trans.__('Time zone');
 
+  let cronString;
+  try {
+    if (props.schedule !== undefined && !props.errors['schedule']) {
+      cronString = cronstrue.toString(props.schedule);
+    }
+  }
+  catch (e) {
+    // Do nothing; let the errors or nothing display instead
```
dlqqq, line 35: Why not console error this?

dlqqq (changes requested):

src/components/schedule-inputs.tsx
jweill-aws: Because it's not an error; I have to have the `try`/`catch` block because, if a cron expression is not valid (and `''` is not valid), `cronstrue` throws an error that breaks the entire page, when really we need to display the error in `props.errors` or nothing at all (e.g., for initial load)

andrii-i (approved):
LGTM.  This is good, greatly improves usability

src/components/schedule-inputs.tsx
jweill-aws: I like the idea of having only one dependency; I can see after our 10:00 meeting whether I can use `cronstrue` to capture errors.

`cron` has a lot of edge cases; for example, `0 5 30 2 *` (at 05:00 every year on February 30) technically passes regexes, but because that date is impossible, a job with this schedule will never run.

dlqqq (approved):

PR #104, opened 2022-10-06
remove idempotency token from all job definition models and API requests

dlqqq:
Idempotency tokens are not valid for job definitions. This removes all erroneous occurrences of it in data models and API requests.


3coins (approved):
Thanks for making these updates. 🚀

PR #106, opened 2022-10-07
Added task runner to create scheduled jobs from job definition

3coins:


jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+                    job_definition_id=model.job_definition_id, next_run_time=next_run_time
+                )
+            )
+        elif active_changed and not active:
+            self.cache.update(model.job_definition_id, UpdateJobDefinitionCache(active=active))
+        elif schedule_changed or timezone_changed:
+            self.cache.update(
+                model.job_definition_id,
+                UpdateJobDefinitionCache(timezone=timezone, schedule=schedule),
+            )
```
dlqqq, line 240: I'm not sure why this method is so overly verbose. I suggest we rewrite this. Here's the flow the update path should take:

1. When we update a job definition, we **always** want to update the cache (i.e. it is a write-through cache). We do not need or want to try to be intelligent about when we're doing so, since that opens us up to possible cache inconsistency. For example, what if a new field was needed by the task runner and needed to be added to the job definition cache? If you forgot to update this specific method, you would have cache inconsistency, because you're only updating the cache if any of the previous cache fields changed.

2. To determine whether we add this job definition into the queue, we just need to check if the previous `next_run_time` and new `next_run_time` differ, and if the new job definition is active.

Rough draft for what you would want here:

```py
cache = self.cache.get(model.job_definition_id)
next_run_time = self.compute_next_run_time(model.schedule, model.timezone)

# always update write-through cache
self.cache.update(model.job_definition_id, UpdateJobDefinitionCache(**model))

# if next run time differs, and new job definition is active, add it to the queue
if cache.next_run_time != next_run_time and model.active:
  self.queue.push(JobDefinitionTask(job_definition_id=model.job_definition_id, next_run_time=next_run_time))
```

jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def delete_job_definition(self):
+        pass
```
dlqqq, line 243: Should delete the record from the cache.

jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+        )
+        self.queue.push(
+            JobDefinitionTask(
+                job_definition_id=definition.job_definition_id, next_run_time=next_run_time
+            )
+        )
+
+    def create_job(self, job_definition_id: str):
+        definition = self.scheduler.get_job_definition(job_definition_id)
+        if definition and definition.active:
```
dlqqq, line 262: Get the job definition from the cache (in memory) instead of the scheduler (in disk).

jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+            cache = self.cache.get(task.job_definition_id)
+            cache_run_time = cache.next_run_time
+            queue_run_time = task.next_run_time
+
+            if not cache.active or queue_run_time != cache_run_time:
+                self.queue.pop()
+                continue
+
+            time_diff = self.compute_time_diff(queue_run_time, cache.timezone)
+            print(f"time difference is : {(time_diff)/1000}")
```
dlqqq, line 283: For clarity, I suggest we make the following changes:

1. When comparing times, I suggest we always convert them to UTC first for clarity. This logic converts the current time to the task's specified time zone, which I think is much more confusing (although equivalent). This means we can delete `get_localized_timestamp()` and instead just convert `task.next_run_time` to UTC.
2. It's very confusing that the default timezone (UTC) is set by `TaskRunner.compute_time_diff()` in a ternary statement. When setting default values for records, we should set them as "high up" as possible for visibility. I suggest making job definition timezones non-optional and enforcing the default value in either the `JobDefinition` table model or in the `CreateJobDefinition` request model.

jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+
+            # if run time is in future
+            if time_diff < 0:
+                break
+            # if run time is in the past might be in previous run
+            elif time_diff >= (self.run_interval * 1000):
+                break
```
dlqqq, line 290: I don't understand this block. Why are we halting execution of jobs when we encounter a job that should be run but is within one task runner interval of the current time? I suspect this is a workaround for some runtime bug rather than intended functionality.

dlqqq (changes requested):

jupyter_scheduler/task_runner.py
3coins: This is a good point, but something to note here is that `model` here might not have any of the 3 values we are interested in here. Also, relying on the `next_run_time` for comparison means we might be pushing the next set of tasks to the queue even if  the `schedule` or `timezone` hasn't changed; I am not sure this comparison will work for all cron expressions.

jupyter_scheduler/task_runner.py
3coins: Cache doesn't have all attributes to create the job.

jupyter_scheduler/task_runner.py
3coins: My initial thought process was avoiding running tasks that were too much in the past because of any runtime situation. But re-visiting this, I am not sure if this condition will ever occur, may be this will happen more if the the run interval is smaller. Since, we are re-hydrate the cache/queue at the start of the server, any past tasks should not appear.

jupyter_scheduler/task_runner.py
3coins: For 1), I am not sure what you mean, and if that will work with the lib `croniter`, we are using here. I spent quite some time making sure this method works with different cron expressions and for timezones; we should add some tests to make sure this doesn't regress before making changes here.

2) is reasonable, does UTC have an equivalent named timezone?

jupyter_scheduler/scheduler.py
```diff
@@ -272,15 +283,21 @@ def create_job_definition(self, model: CreateJobDefinition) -> str:
...
+            self.task_runner.add_job_definition(job_definition_id)
+
         return job_definition_id
 
     def update_job_definition(self, model: UpdateJobDefinition):
         with self.db_session() as session:
             session.query(JobDefinition).filter(
                 JobDefinition.job_definition_id == model.job_definition_id
-            ).update(model.dict(exclude_none=True))
+            ).update(model.dict(exclude_none=True, exclude={"job_definition_id"}))
```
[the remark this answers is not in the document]
3coins: I don't recall, but I think I faced some errors.

jupyter_scheduler/models.py
```diff
@@ -192,9 +192,10 @@ class UpdateJobDefinition(BaseModel):
     tags: Optional[Tags] = None
     name: Optional[str] = None
     url: Optional[str] = None
+    schedule: Optional[str] = None
     timezone: Optional[str] = None  # Should be a timezone e.g., US/Eastern, Asia/Kolkata
     output_filename_template: Optional[str] = OUTPUT_FILENAME_TEMPLATE
-    active: bool
+    active: Optional[bool] = None
```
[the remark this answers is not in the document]
3coins: It is reasonable to have a job definition without schedule because, it might be used for creating run now jobs as well. I can fix this error by not invoking task runner when there is no schedule present.

jupyter_scheduler/scheduler.py
```diff
@@ -133,9 +133,20 @@ def resume_jobs(self, job_definition_id: str):
...
     _db_session = None
+    _task_runner = None
 
-    def __init__(self, config: ExecutionConfig = {}):
+    def __init__(self, config: ExecutionConfig = {}, task_runner_class=None):
         self.config = config
+        self.task_runner_class = task_runner_class
+
+    @property
+    def task_runner(self):
```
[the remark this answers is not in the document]
3coins: I was getting some null pointer exceptions without this. Have to check if code will work with direct assignments.

jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+
+    @property
+    def session(self):
+        if not self._session:
+            self.cache_url = "sqlite://"
+            engine = create_engine(self.cache_url, echo=False)
+            Base.metadata.create_all(engine)
+            self._session = sessionmaker(bind=engine)
+
+        return self._session
```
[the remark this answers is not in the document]
3coins: Ditto as above.

jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+            self._cache = Cache()
+
+        return self._cache
+
+    @property
+    def queue(self):
+        if not self._queue:
+            self._queue = PriorityQueue()
+
+        return self._queue
```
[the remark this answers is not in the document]
3coins: Same as above.

jupyter_scheduler/task_runner.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def compute_next_run_time(self, schedule: str, timezone: Optional[str] = None):
+        return compute_next_run_time(schedule, timezone)
```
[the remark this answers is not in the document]
3coins: This pattern is useful because it localizes change to one place if something changes, for example if you replaced the inner function with a different one. It is also useful for mocking the method for testing rather than patching the method from utils.

jupyter_scheduler/task_runner.py
dlqqq: Ah, you're right. And it shouldn't be a big deal since we're already doing a disk write when creating a new job.

jupyter_scheduler/task_runner.py
dlqqq: 1. I'm saying I personally find this implementation confusing. For time comparisons, I would have expected the job time to be converted to the server's time zone, or for both of them to be converted to UTC. Not sure how else to rephrase this. Because your implementation is equivalent and my concerns are developer-facing, I won't block it, but I do urge your consideration.

2. `Etc/UTC`.

jupyter_scheduler/task_runner.py
dlqqq: My problem with this logic is that you're not popping the item from the heap. If the heap somehow ends up in an invalid state where the top of the heap should have already been run (and thus popped from the heap), the task runner is completely blocked and no new scheduled jobs are run. With the current logic, this exact same condition will be true for all subsequent iterations.

> But re-visiting this, I am not sure if this condition will ever occur, may be this will happen more if the run interval is smaller (1s - 2s). Since, we are re-hydrate the cache/queue at the start of the server, any past tasks should not appear.

Yeah, the cache gets dropped on server shutdown and `next_run_time` is only calculated relative to the current time. Therefore, this condition shouldn't happen to begin with.

I do want to keep this conversation open though, since I did want to call into question whether we should be running past tasks upon server start. For example, if I had a schedule to run on Wednesdays and shut down my server that week, when I start my server again the following Sunday, should it start running my job from last Wednesday? I think we should have a discussion on this.

jupyter_scheduler/task_runner.py
dlqqq: Oh right, the update model is a partial of the job definition model. My bad, you do actually want to include the below logic in your implementation:

```
        schedule = model.schedule or cache.schedule
        timezone = model.timezone or cache.timezone
```

However, there is another issue here that's inherent to how we are handling update API requests. Because `model.schedule` allows `None` for a value, it seems impossible to distinguish from the `UpdateJobDefinition` model whether the user left the field unspecified, or if they specified the field to be `None` (e.g. the difference in JSON request bodies is `{}` v.s. `{ schedule: null }`). In both cases, `model.schedule` is set to `None`.

> Also, relying on the next_run_time for comparison means we might be pushing the next set of tasks to the queue even if the schedule or timezone hasn't changed; I am not sure this comparison will work for all cron expressions.

This should never happen, since if `schedule` and `timezone` have not changed, then `next_run_time` should not change either.

jupyter_scheduler/models.py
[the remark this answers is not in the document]
dlqqq: Interesting, so job definitions can be used as both a template for "run now" and as a schedule. I'm OK with this as long as it was already agreed upon by our team.

jupyter_scheduler/task_runner.py
3coins: Let's discuss this and iterate in another PR; I have other questions around this.

jupyter_scheduler/extension.py
```diff
@@ -87,10 +95,12 @@ def initialize_settings(self):
...
 
-        scheduler = self.scheduler_class(config=execution_config, task_runner_class=TaskRunner)
+        scheduler = self.scheduler_class(
+            config=execution_config, task_runner_class=self.task_runner_class
+        )
 
         self.settings.update(execution_config=execution_config, scheduler=scheduler)
 
-        if scheduler.task_runner:
+        if self.task_runner_run_interval > -1 and scheduler.task_runner:
```
dlqqq, line 104: Well, this condition still allows for an invalid value like `-0.9`. Probably should be `<= 0`, since `0` also doesn't seem valid.

dlqqq (changes requested):

jupyter_scheduler/task_runner.py
dlqqq: The new logic looks great. 👍

dlqqq: Tracking in #119.

jupyter_scheduler/scheduler.py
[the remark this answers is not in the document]
3coins: Updated.

jupyter_scheduler/task_runner.py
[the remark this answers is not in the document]
3coins: Updated.

jupyter_scheduler/task_runner.py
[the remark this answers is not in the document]
3coins: Updated.

dlqqq (approved):
Looks good. All other changes can be addressed in follow-up PRs.

PR #115, opened 2022-10-10
Sets default time zone in create-job form

### LINKED ISSUE #112: Automatic time zone detection
### Problem

The time zone is blank by default, and the user must select one every time.

### Proposed Solution

Attempt to guess the user's time zone and prepopulate the "Time zone" picker in the create form with a best guess.

### Additional context

Browsers can report their own time zone in the form of a UTC offset: for example, on 2022-10-10, Seattle is in UTC-07:00. A UTC offset by itself is not a time zone; for example, Phoenix is also currently in UTC-07:00, but it does not observe daylight saying time, and so Phoenix will remain at UTC-07:00 in mid-November, when Seattle is in UTC-08:00.

jweill-aws:
Uses the browser API, supported in all browsers (with fallback to `undefined`), to retrieve the default time zone for the user. Prepopulates it in forms that could create a new job definition.

Fixes #112.

https://user-images.githubusercontent.com/93281816/194961306-23421cb9-1da0-4936-aaf0-5e8d6511e0f1.mov

### DIFF
--- a/src/components/job-row.tsx
+++ b/src/components/job-row.tsx
@@ -93,7 +93,8 @@ function RefillButton(props: {
       environment: props.job.runtime_environment_name,
       parameters: jobParameters,
       createType: 'Job',
-      scheduleInterval: 'weekday'
+      scheduleInterval: 'weekday',
+      timezone: Intl.DateTimeFormat().resolvedOptions().timeZone
     };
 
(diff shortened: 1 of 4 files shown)


andrii-i (approved):
Works on latest Chrome, Safari, Firefox

PR #131, opened 2022-10-12
Made update apis consistent with REST APIs

### LINKED ISSUE #128: Scheduler update apis are not consistent with REST APIs
### Problem
Scheduler update APIs are using models that require `job_id` or `job_definition_id` as required attributes. This means, that the payload from the REST call has to include these. This is inconsistent with the REST APIs, as the ids should be populated from the path query. 

### Proposed Solution

- Remove `job_id` from `UpdateJob` model
- Update definition of `update_job` in `Scheduler.py` to take `job_id` as one of the inputs
- Remove `job_definition_id` from `UpdateJobDefinition` model
- Update definition of `update_job_definition` in `Scheduler.py` to take `job_definition_id` as one of the inputs.

### Additional context

https://github.com/jupyter-server/jupyter-scheduler/blob/9b087034c35fb332104f8c45c191370089c0a1f3/jupyter_scheduler/models.py#L146-L152

https://github.com/jupyter-server/jupyter-scheduler/blob/9b087034c35fb332104f8c45c191370089c0a1f3/jupyter_scheduler/models.py#L184-L198

https://github.com/jupyter-server/jupyter-scheduler/blob/9b087034c35fb332104f8c45c191370089c0a1f3/jupyter_scheduler/scheduler.py#L49

https://github.com/jupyter-server/jupyter-scheduler/blob/9b087034c35fb332104f8c45c191370089c0a1f3/jupyter_scheduler/scheduler.py#L99

3coins:
Fixes #128

### DIFF
--- a/jupyter_scheduler/handlers.py
+++ b/jupyter_scheduler/handlers.py
@@ -100,8 +100,9 @@ class JobDefinitionHandler(ExtensionHandlerMixin, JobHandlersMixin, APIHandler):
     @tornado.web.authenticated
     async def patch(self, job_definition_id):
         payload = self.get_json_body()
-        payload["job_definition_id"] = job_definition_id
-        await ensure_async(self.scheduler.update_job_definition(UpdateJobDefinition(**payload)))
+        await ensure_async(
+            self.scheduler.update_job_definition(job_definition_id, UpdateJobDefinition(**payload))
+        )
         self.set_status(204)
         self.finish()
 
@@ -163,16 +164,16 @@ class JobHandler(ExtensionHandlerMixin, JobHandlersMixin, APIHandler):
     async def patch(self, job_id):
         payload = self.get_json_body()
 
-        if "status" not in payload:
-            raise tornado.web.HTTPError(500, "Field 'status' missing in request body")
+        status = payload.get("status", None)
+        status = Status(status) if status else None
 
-        status = Status(payload.get("status"))
-        if status == Status.STOPPED:
+        if status and status != Status.STOPPED:
+            raise tornado.web.HTTPError(500, "Value of 'STOPPED' only allowed for field 'status'")
+
+        if status:
             await ensure_async(self.scheduler.stop_job(job_id))
         else:
-            await ensure_async(
-                self.scheduler.update_job(UpdateJob(job_id=job_id, status=str(status)))
-            )
+            await ensure_async(self.scheduler.update_job(job_id, UpdateJob(**payload)))
 
         self.set_status(204)
         self.finish()
--- a/jupyter_scheduler/handlers.py
+++ b/jupyter_scheduler/handlers.py
@@ -168,7 +168,10 @@ class JobHandler(ExtensionHandlerMixin, JobHandlersMixin, APIHandler):
         status = Status(status) if status else None
 
         if status and status != Status.STOPPED:
-            raise tornado.web.HTTPError(500, "Value of 'STOPPED' only allowed for field 'status'")
+            raise tornado.web.HTTPError(
+                500,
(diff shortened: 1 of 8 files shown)


jupyter_scheduler/handlers.py
```diff
@@ -163,16 +164,16 @@ async def post(self):
     async def patch(self, job_id):
         payload = self.get_json_body()
 
-        if "status" not in payload:
-            raise tornado.web.HTTPError(500, "Field 'status' missing in request body")
+        status = payload.get("status", None)
+        status = Status(status) if status else None
+
+        if status and status != Status.STOPPED:
+            raise tornado.web.HTTPError(500, "Value of 'STOPPED' only allowed for field 'status'")
```
dlqqq, line 171: 1. I'd rename the status message to something like "Invalid value for field 'status'. Jobs can only be updated to status 'STOPPED' after creation."

2. Doesn't this technically allow you to stop a completed job? It looks like you're setting the status to 'STOPPED' as long as it's in the request, regardless of the current job's status. IMO, stopping a job only makes sense if the job is 'QUEUED' or 'IN PROGRESS'. If you do make this change, add a new corresponding error message as well, such as: "Invalid value for field 'status'. Only jobs that are either 'QUEUED' or 'IN PROGRESS' can be stopped."

dlqqq (changes requested):

jupyter_scheduler/handlers.py
jweill-aws, line 171: This is confusing; it suggests that you've supplied a value 'STOPPED' for a field that is not 'status'

```suggestion
            raise tornado.web.HTTPError(500, "Field 'status' can be set to 'STOPPED' only")
```

jupyter_scheduler/tests/test_handlers.py
```diff
@@ -162,11 +163,34 @@ async def test_get_jobs(jp_fetch, params, list_query, jobs_list):
...
+
+
+async def test_patch_jobs_for_invalid_status(jp_fetch):
     with pytest.raises(HTTPClientError) as e:
         job_id = "542e0fac-1274-4a78-8340-a850bdb559c8"
-        response = await jp_fetch("scheduler", "jobs", job_id, method="PATCH", body="{}")
-    assert expected_http_error(e, 500, "Field 'status' missing in request body")
+        body = {"status": "IN_PROGRESS"}
+        await jp_fetch("scheduler", "jobs", job_id, method="PATCH", body=json.dumps(body))
+        assert expected_http_error(e, 500, "Value of 'STOPPED' only allowed for field 'status'")
```
jweill-aws, line 182: Change to match revised error message

3coins:
cc @dylanramzn

jupyter_scheduler/handlers.py
3coins: For 2), No, the scheduler API does not allow to stop a job.

jupyter_scheduler/models.py
```diff
@@ -193,9 +189,10 @@ class UpdateJobDefinition(BaseModel):
     name: Optional[str] = None
     url: Optional[str] = None
     schedule: Optional[str] = None
-    timezone: Optional[str] = None  # Should be a timezone e.g., US/Eastern, Asia/Kolkata
+    timezone: Optional[str] = None
```
jweill-aws, line 192: I like the idea of adding that this should be in tzdata format (although `US/Eastern` is an alias for `America/New_York`).

jweill-aws (approved):

jupyter_scheduler/models.py
3coins: I will add those in the docs; haven't figured out the mechanism to generate them yet.

PR #141, opened 2022-10-12
Job descriptions list actions: pause/resume/delete

### LINKED ISSUE #97: Add "Paused" column, pause button to Job Definitions list
### Problem

Users want to see whether a job definition is active or paused. Users want to pause or unpause a job definition.

### Proposed Solution

- [x] Add a "Paused" column to the Job Definitions List view. If a job definition is not "active," it is paused.
- [ ] Add a "pause" button to pause an active job or an "unpause" button to make a paused job active.

### Additional context

See #81.

jweill-aws:
Fixes #97.

Adds pause, resume, and delete buttons using MUI icons to the job definitions list.

For consistency, uses MUI icons and MUI IconButtons in the jobs list as well.

Forces a reload after pause and resume.

https://user-images.githubusercontent.com/93281816/195463784-367c331a-e769-45bb-83dc-350ba4fb2a3c.mov


src/components/job-definition-row.tsx
```diff
@@ -34,13 +39,78 @@ function ScheduleSummary(props: {
...
-  openJobDefinitionDetail: (jobDefId: string) => unknown
+  openJobDefinitionDetail: (jobDefId: string) => unknown,
+  deleteRow: (
+    id: Scheduler.IDescribeJobDefinition['job_definition_id']
+  ) => void,
+  forceReload: () => void
 ): JSX.Element {
   const trans = useTranslator('jupyterlab');
 
+  const ss = new SchedulerService({});
```
dlqqq, line 112: 1. It's invalid to call a React hook outside of a React component. Although the type signature of this function is identical (e.g. `(...args: any[]) => JSX.Element`), this is still not being used like a stateful React component. The translator really should be passed in by the parent component as an argument. 

2. Because this function is called every render, the scheduler service instance is also being initialized every render. Ideally the scheduler service instance would be passed to the builder as an argument as well.

I know that the "row builder" pattern is ugly and leads to the "parameter overload" we're seeing; I had intended to refactor this later but more pressing concerns emerged. Keep in mind that there's nothing really wrong with this (i.e. it doesn't significantly increase our risk of runtime bugs) other than the fact it's ugly, so the refactor is pretty low priority ATM.

dlqqq (changes requested):

src/components/job-definition-row.tsx
jweill-aws: I had previously wrapped this in a `useMemo` hook, but the whole page rendered as blank and I got a `Rendered more hooks than during the previous render` error. I can make the instances be parameters instead.

dlqqq (approved):

PR #153, opened 2022-10-14
Advanced options errors

### LINKED ISSUE #101: Badge advanced options when at least one input has an error
### Problem

If the advanced options panel is collapsed, and at least one input in it has an error, the user might not know whey they cannot submit the form.

### Proposed Solution

When the expandable section is collapsed and at least one error is contained in it, display an indicator on the section.

Alternatively, keep the section expanded and do not let the user collapse the section if at least one error is present.

jweill-aws:
Fixes #101. If at least one input in the advanced options section has an error, mark the "Advanced options" header with an error icon to indicate that some input in this section has an error, even if the section is collapsed.

Demo video:

https://user-images.githubusercontent.com/93281816/195898814-0411f0a2-bc7c-4bfe-a50b-560c475c1c49.mov

Screen shot:

image (5)

### DIFF
--- a/src/mainviews/create-job.tsx
+++ b/src/mainviews/create-job.tsx
@@ -15,6 +15,8 @@ import { useTranslator } from '../hooks';
 import { ICreateJobModel, IJobParameter, ListJobsView } from '../model';
 import { Scheduler as SchedulerTokens } from '../tokens';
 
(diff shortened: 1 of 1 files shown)


awaisabir (approved):
This should be a non-intrusive change! Thanks for this 😄

PR #155, opened 2022-10-14
Handle create error in create form

jweill-aws:
Partial fix for #147. If the `createJob` or `createJobDefinition` request results in a rejection for some reason, the user remains on the create form and an error alert is displayed on top.

After the user attempts to resubmit the form or modify one of the inputs, the error alert is cleared.

Also runs prettier on all files in repo.

Behavior when the `createJob` handler is overridden to return an error all the time:

https://user-images.githubusercontent.com/93281816/195953166-cf84ad98-05cc-4fd8-8aac-a1c21f8bb20a.mov

### DIFF
--- a/src/mainviews/create-job.tsx
+++ b/src/mainviews/create-job.tsx
@@ -103,20 +103,17 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
     const parameterValueIdx = parameterValueMatch(target.name);
     const newParams = props.model.parameters || [];
 
-    let newModel = { ...props.model, createError: undefined };
     if (parameterNameIdx !== null) {
       newParams[parameterNameIdx].name = target.value;
-      newModel.parameters = newParams;
+      props.handleModelChange({ ...props.model, parameters: newParams });
     } else if (parameterValueIdx !== null) {
       newParams[parameterValueIdx].value = target.value;
-      newModel.parameters = newParams;
+      props.handleModelChange({ ...props.model, parameters: newParams });
     } else {
       const value = target.type === 'checkbox' ? target.checked : target.value;
       const name = target.name;
-      newModel = { ...newModel, [name]: value };
+      props.handleModelChange({ ...props.model, [name]: value });
     }
-
-    props.handleModelChange(newModel);
   };
 
   const validateSchedule = (schedule: string) => {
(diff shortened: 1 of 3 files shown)


andrii-i (approved):
This is nice

src/mainviews/create-job.tsx
```diff
@@ -576,10 +577,15 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+    api
+      .createJob(jobOptions)
+      .then(response => {
+        // Switch to the list view with "Job List" active
+        props.showListView('Job');
+      })
+      .catch((error: string) => {
+        props.handleModelChange({ ...props.model, createError: error });
+      });
   };
```
dlqqq, line 589: Believe we want to briefly immediately `createError` upon form submission. This is consistent with the UX in most other forms I've used. If the user gets a create error, and then retries to submit the form, this behavior provides additional visual emphasis on the fact that another create error happened, rather than having the user think that the form's buggy. To do this, just append this snippet below line 589:

```
props.handleModelChange(model => ({ ...model, createError: undefined }));
```

dlqqq (changes requested):

src/mainviews/create-job.tsx
```diff
@@ -642,26 +677,37 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+      createError: undefined,
+      parameters: newParams
+    });
   };
 
   const formPrefix = 'jp-create-job-';
 
   const cantSubmit = trans.__('One or more of the fields has an error.');
+  const createError: string | undefined = props.model.createError;
+  console.log('createError is ' + createError);
```
dlqqq: ❌

dlqqq (changes requested):
I only meant that errors should be cleared when clicking the submit button, but I'm open to experimenting with the UX of clearing errors immediately on any input change. Two issues I foresee:

1. It's easy for future developers to forget to set `createError: undefined` when handling change events. Especially if they're using an advanced options form extension and aren't cross-referencing their code with the implementation here. This would mean some input fields clear the error while others don't, which can be confusing to the user.

2. Users may want to keep the error visible while editing their form. It's easy to see how a user might immediately rush to change their form and then forget what the original error was.

dlqqq (approved):

PR #157, opened 2022-10-15
Updated id regex to allow non-uuid ids

### LINKED ISSUE #145: Relax id requirement for jobs and job definitions to be any alphanumeric value
### Problem
Current REST endpoints require the job and job_definition ids to be specifically `uuid` format. This is restrictive in nature, as certain backend implementations might have different id formats.

### Proposed Solution
Relax the id requirement in REST endpoints to allow any alphanumeric value.

### Additional context

https://github.com/jupyter-server/jupyter-scheduler/blob/8c1deaf22360414b71f72c6acb5a897f62d11238/jupyter_scheduler/extension.py#L25-L39

### LINKED ISSUE #147: Handle errors in server and UI at job creation time

3coins:
Fixes #145 
Partially fixes #147

### DIFF
--- a/jupyter_scheduler/exceptions.py
+++ b/jupyter_scheduler/exceptions.py
@@ -1,0 +1,14 @@
+class InputUriError(Exception):
+    def __init__(self, input_uri: str):
+        self.input_uri = input_uri
+
+    def __str__(self):
+        return f"Input path '{self.input_uri}' does not exist."
+
+
+class IdempotencyTokenError(Exception):
+    def __init__(self, idempotency_token: str):
+        self.idempotency_token = idempotency_token
+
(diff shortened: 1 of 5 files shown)


3coins:
cc @bhadrip

ellisonbg (approved):

PR #164, opened 2022-10-18
Expose Advanced Options in Job Definition Detail 

### LINKED ISSUE #160: Expose Advanced Options in Job Definition Detail
This issue needs validation

## Description

Currently `Advanced Options` are only shown at the bottom of `Job Detail` page. At the same time, as mentioned by @3coins, fields provided to `Advanced Options` might be / are meaningful for `Job Definition Detail` page too, especially if `Job Definition` can be edited as user might need to edit `Advanced Options` for `Job Definition` too

## Expected behavior

Needs confirmation: `Advanced Options` are shown in Job Definition Detail

andrii-i:
Fixes #160 

## Description

This PR exposes and renders `AdvancedOptions` in `JobDefinitionDetail`. To achieve that, `AdditionalOptions` now also optionally accept `IJobDefinitionModel` in addition to `ICreateJobModel | IJobDetailModel` as a `model`. To minimize changes to how `AdvancedOptions` already work in `Job Detail`, I added 'AdvancedOptions' to `JobDefinitionDetail` vs pulling it up into `JobDetailView`

## Issue Description
Currently Advanced Options are only shown at the bottom of Job Detail page. At the same time, as mentioned by @3coins, fields provided to Advanced Options might be / are meaningful for Job Definition Detail page too, especially if Job Definition can be edited as user might need to edit Advanced Options for Job Definition too

## Expected behavior
Needs confirmation: Advanced Options are available and shown in Job Definition Detail

## Preview
jpb_def_adv_options_below_list

### DIFF
--- a/src/mainviews/detail-view/detail-view.tsx
+++ b/src/mainviews/detail-view/detail-view.tsx
@@ -141,7 +141,6 @@ export function DetailView(props: IDetailViewProps): JSX.Element {
               setListJobsView={props.setListJobsView}
               // Extension point: optional additional component
               advancedOptions={props.advancedOptions}
-              outputFormatsStrings={outputFormatStrings ?? []}
             />
           )}
           {props.model.detailType === 'JobDefinition' && jobDefinitionModel && (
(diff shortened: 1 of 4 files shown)


dlqqq (approved):
Hmm, an obvious issue with this is that it appears the advanced options inputs will still be editable in the detail view. We should add a `readonly` boolean prop to the existing advanced options API to address this, but there's no need to do so at this very moment.

andrii-i:

Good point. Created #165 to track it (also part of / mentioned in #87)

PR #167, opened 2022-10-19
Fix for create job UI when error is returned from backend

3coins:
Current create job UI is unstable if an error is returned from backend after form submit. The cause of this error was incorrect type in the error handler code. This PR fixes this code, so that correct error message is passed on to the model for rendering.


jweill-aws (approved):
Thanks for fixing this!

PR #189, opened 2022-10-23
Added jobs directory to store input copy and outputs

### LINKED ISSUE #162: store absolute paths for output files in DB
### Problem

Storing job output directories as relative to the server root fails if the server is shut down and then started in a different directory.

### Proposed Solution

- when building the output directory path in the backend, store it in the DB as an absolute path
- when returning output file paths from the backend, always have it be relative to the server root

However, it's not clear how the backend would handle output file paths from entirely disjoint directories. Take the following directory structure:

```
/home/user/
|_ a/
|_ b/
```

If a job were created when `a/` was the server root, then when `b/` is the server root, how are those output paths returned from the backend? Files under `a/` are inaccessible to the ContentsManager and cannot be viewed via the JL UI when served from `b/`. I believe that in this case, the backend should just return the absolute file path and the frontend can handle the UI accordingly, via determining if the path is absolute or not. Perhaps greying out the output files link and then rendering a tooltip (cannot view files outside server root)?

3coins:
## Description

- removes "re-run job" functionality
- removes "input URI", replaced with "input filename"
- renames "output files" to "job files"
- copies input files to staging area upon creation of job & job definition
- execution manager reads input file from staging area rather than user's original input file under server root
- input files are included when downloading job files

## Issues closed

- Closes #162
- Closes https://github.com/jupyter-server/jupyter-scheduler/issues/192


3coins:
Added unit tests for job files manager.

andrii-i:
Fixes #192

PR #197, opened 2022-10-24
Adds default compute type to runtime env model, uses it

### LINKED ISSUE #188: Default compute type per environment
### Problem

Currently, we default to the first compute type per environment (#177). We may want to enable an environment to provide a default compute type along with the list of supported compute type values.

### Proposed Solution

Modify the API return type to allow not only a list of compute types, but also a preferred compute type.

In the UI, if no compute type is yet selected (i.e., the user is creating a job and has just selected the environment), by default, display the preferred compute type.

jweill-aws:
Fixes #188. Modifies the backend Python code to add a `default_compute_type` compute type to the `RuntimeEnvironment` model. If a `default_compute_type` is present in the model returned by the Python backend, the front-end code will default to it when an environment is selected. If no default compute type is present in the model, the UI will continue to select the first compute type (see #187).

In the demo below, a custom handler creates three compute types, suffixed `-a`, `-b`, and `-c`, in that order. The default compute type ends in `-b`.

https://user-images.githubusercontent.com/93281816/197642760-53f186e3-9a34-4a1e-bdbd-719e3541a246.mov

### DIFF
--- a/src/mainviews/create-job.tsx
+++ b/src/mainviews/create-job.tsx
@@ -96,7 +96,7 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
         if (
           envList[0].default_compute_type &&
           envList[0].compute_types &&
-          envList[0].default_compute_type in envList[0].compute_types
+          envList[0].compute_types.indexOf(envList[0].default_compute_type) >= 0
         ) {
           newComputeType = envList[0].default_compute_type;
         }
@@ -181,7 +181,7 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
         envObj &&
         envObj.default_compute_type &&
         envObj.compute_types &&
-        envObj.default_compute_type in envObj.compute_types
+        envObj.compute_types.indexOf(envObj.default_compute_type) >= 0
       ) {
         newComputeType = envObj.default_compute_type;
       }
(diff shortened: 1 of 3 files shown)


src/mainviews/create-job.tsx
```diff
@@ -89,10 +89,22 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
       setEnvironmentList(envList);
       if (envList.length === 1) {
+        // If no default compute type is specified, show the first one by default
+        let newComputeType = envList[0].compute_types?.[0];
+
+        // Validate that the default compute type is in fact in the list
+        if (
+          envList[0].default_compute_type &&
+          envList[0].compute_types &&
+          envList[0].compute_types.indexOf(envList[0].default_compute_type) >= 0
```
andrii-i, line 99: `envList[0].compute_types.includes(envList[0].default_compute_type)` could be used here. It would save just 3 characters (no >= 0 needed) but might improve readability. At your discretion

src/mainviews/create-job.tsx
```diff
@@ -160,13 +172,23 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
-    // if setting the environment, default the compute type to its first value (if any are present)
+    // if setting the environment, default the compute type to its default value or its first value
     if (target.name === 'environment') {
       const envObj = environmentList.find(env => env.name === target.value);
+      // Validate that the default compute type is in fact in the list
+      let newComputeType = envObj?.compute_types?.[0];
+      if (
+        envObj &&
+        envObj.default_compute_type &&
+        envObj.compute_types &&
```
andrii-i, line 183: As an option can be shortened to:
```
        envObj?.default_compute_type &&
        envObj?.compute_types &&
```

andrii-i (approved):
LGTM, left some minor code optimization suggestions to be taken at your discretion

Issue #196, opened 2022-10-24
Select all output formats by default


jweill-aws:
### Problem

When creating a new job or job definition, when selecting an environment, the user must select output formats. A default set of output formats, `["ipynb"]`, is used if a user selects none.

### Proposed Solution

Select all output formats by default. If a user selects no output formats, run the notebook and do not specify any output formats.

PR #198, opened 2022-10-24
Select all output formats by default

andrii-i:
## Description
Takes care of "select all output formats by default" part of  #196 
"Allow for selection of no formats" would be taken care of in a separate PR 

## Preview
out_form

### DIFF
--- a/src/mainviews/create-job.tsx
+++ b/src/mainviews/create-job.tsx
@@ -91,6 +91,7 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
     const setList = async () => {
       const envList = await api.getRuntimeEnvironments();
       setEnvironmentList(envList);
+      let environment = props.model.environment;
       if (envList.length === 1) {
         // If no default compute type is specified, show the first one by default
         let newComputeType = envList[0].compute_types?.[0];
@@ -108,6 +109,14 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
           ...props.model,
           environment: envList[0].name,
           computeType: newComputeType
+        });
+        environment = envList[0].name;
+      }
+      const outputFormats = outputFormatsForEnvironment(envList, environment);
+      if (outputFormats) {
+        props.handleModelChange({
+          ...props.model,
+          outputFormats: outputFormats.map(format => format.name)
         });
       }
     };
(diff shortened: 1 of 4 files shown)


src/mainviews/create-job.tsx
```diff
@@ -101,10 +105,26 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+        )?.map(format => format.name);
+
         props.handleModelChange({
           ...props.model,
           environment: envList[0].name,
-          computeType: newComputeType
+          computeType: newComputeType,
+          outputFormats: outputFormats
+        });
+        environment = envList[0].name;
```
dlqqq, line 116: This assignment is unnecessary since `environment` doesn't get used after this. Remove the declaration on line 94 and delete this line.

src/mainviews/create-job.tsx
```diff
@@ -101,10 +105,26 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+      } else {
+        const outputFormats = outputFormatsForEnvironment(
+          envList,
+          environment
+        )?.map(format => format.name);
```
dlqqq, line 121: You don't need to set `outputFormats` here in the `else` case since `props.model.environment` is an empty string. We don't select a default environment in the case where there is more than one environment supported.

src/mainviews/create-job.tsx
```diff
@@ -184,10 +201,15 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+      const newEnvOutputFormats = outputFormatsForEnvironment(
+        environmentList,
+        target.value
+      )?.map(format => format.name);
       props.handleModelChange({
         ...props.model,
         environment: target.value,
-        computeType: newComputeType
+        computeType: newComputeType,
+        outputFormats: newEnvOutputFormats
```
dlqqq, line 212: The problem is that the UI is not reflecting the actual state of the model when you change the environment. When selecting a new environment, all output formats are re-enabled in the model, but this is not reflected in the UI.

https://user-images.githubusercontent.com/44106031/197876814-ec6df15c-8528-4149-a827-64655cec2ee0.mov

To remedy this, change the `defaultChecked` prop in `output-format-picker.tsx` to `checked`. `defaultChecked` is only respected at initial render, and should only be used for uncontrolled components.

dlqqq (changes requested):

src/mainviews/create-job.tsx
andrii-i: This makes sense, thanks for explanations

andrii-i:
Updated preview in the PR description

dlqqq (approved):
Love how much shorter the diff is after your revisions. Looks great 👍

andrii-i:
I'm particularly happy that if-else branching in useEffect is not necessary

PR #202, opened 2022-10-25
Support backends without `total_count` in list views

### LINKED ISSUE #182: UI pagination support for backends that doesn't support total count
### Problem
Certain backends will not support a total count returned in their list APIs. The pagination component in this case will be incompatible showing, for example `0-0 of 0` in the counts.
Screen Shot 2022-10-21 at 12 42 55 PM

### Proposed Solution
Enhance the pagination component to stop showing count if a zero is returned. In addition, pagination should only show the next and previous buttons to indicate that there are more pages available.

### Additional context
Version: 0.5.1

dlqqq:
- correctly renders row count label even when `total_count` is not implemented by the backend
- fixes seed script and uses real input files (ported from #175)
- fixes lint

## Expected behavior

- until last page is reached, the total row count should be rendered with a '+' to indicate more rows may be present
- upon reaching the last page, '+' should disappear and the next button should be disabled
- going back after reaching the last page should keep the '+' hidden as we now know the total count

## To reproduce

Run the seed script. It's intended to have JupyterLab served from `dev/`.

```
cd dev/
python seed.py
jupyter lab
```

Then change the body of `SchedulerService.getJobs()` in `handler.ts`:

```
    ...
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
    delete data.total_count;
    return data as Scheduler.IListJobsResponse;
```

## Demo

https://user-images.githubusercontent.com/44106031/197866246-a83454e8-5549-4415-8f78-866043cb668c.mov

## Issues closed

- Closes #182.

### DIFF
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -175,8 +175,8 @@ export function AdvancedTable<
           return trans.__('%1–%2 of %3', from, loadedRows, loadedRows);
         } else {
           return (
-            trans.__('%1–%2 of %3', from, to, loadedRows) +
-            (nextToken === undefined ? '' : '+')
+            trans.__('%1–%2 of %3', from, to,
+              loadedRows + (nextToken === undefined ? '' : '+'))
           );
         }
       } else {
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -1,10 +1,12 @@
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 
 import { useTheme } from '@mui/material/styles';
 import Table from '@mui/material/Table';
 import TableContainer from '@mui/material/TableContainer';
 import TableBody from '@mui/material/TableBody';
-import TablePagination from '@mui/material/TablePagination';
+import TablePagination, {
+  LabelDisplayedRowsArgs
+} from '@mui/material/TablePagination';
 import Paper from '@mui/material/Paper';
 
 import { Scheduler } from '../../handler';
@@ -20,7 +22,7 @@ export type AdvancedTableColumn = {
 type AdvancedTablePayload =
   | {
       next_token?: string;
-      total_count: number;
+      total_count?: number;
     }
   | undefined;
 
@@ -69,7 +71,7 @@ export function AdvancedTable<
 >(props: AdvancedTableProps<P, Q, R>): JSX.Element {
   const [rows, setRows] = useState<R[]>();
   const [nextToken, setNextToken] = useState<string>();
-  const [totalCount, setTotalCount] = useState<number>(0);
+  const [totalCount, setTotalCount] = useState<number>();
   const [page, setPage] = useState<number>(0);
   const [maxPage, setMaxPage] = useState<number>(0);
   const [loading, setLoading] = useState<boolean>(true);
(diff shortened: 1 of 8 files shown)


dlqqq:
Let's not merge this until we get feedback from @ellisonbg.

src/components/advanced-table/advanced-table.tsx
```diff
@@ -153,8 +155,35 @@ export function AdvancedTable<
...
+   */
+  const labelDisplayedRows = useCallback(
+    ({ from, to, count }: LabelDisplayedRowsArgs) => {
+      if (count === -1) {
+        const loadedRows = rows?.length ?? 0;
+        if (onLastPage) {
+          // for some reason `to` is set incorrectly on the last page in
+          // server-side pagination, so we need a custom template literal for
+          // this case.
+          return `${from}-${loadedRows} of ${loadedRows}`;
```
jweill-aws, line 173: This should use `trans.__()` for localization.

src/components/advanced-table/advanced-table.tsx
```diff
@@ -153,8 +155,35 @@ export function AdvancedTable<
...
+        } else {
+          return (
+            `${from}-${to} of ${loadedRows}` +
```
jweill-aws, line 176: This should also use `trans.__()`. I don't think that the "25+", "50+" values are meaningful. The user already knows that there is a next page if the `>` button is active. I think that if the system doesn't provide a total count, we should just display `X–Y` instead of `X–Y of Y+`.

src/components/advanced-table/advanced-table.tsx
dlqqq: I don't think `trans.__()` works for dynamic strings like this. To my recollection, it's implemented naively as a dictionary mapping English strings to their localized counterparts. There's no way that would work with dynamic strings, since you would need a key for every possible string that can be generated from this template literal.

I suppose we could translate the word "of", but that brings its own localization difficulties.

I'd say for now we just leave this as-is for now until we find a better localization implementation that can handle this case.

jweill-aws: `trans_('%1–%2 of %3', from, to, total)`

src/components/advanced-table/advanced-table.tsx
dlqqq: I disagree. It's not clear with just `X-Y` that it's referring to the row count. For example, given a page size of 5 (e.g. in the detail views), it's not immediately obvious that the string `1-5` refers to the rows in my table. The user would have to count the 5 rows in the table and then infer "ah, this is referring to how many rows are being rendered by the table". `1-5 of 5+` makes this more clear.

src/components/advanced-table/advanced-table.tsx
dlqqq: TY! I apologize for not doing more research on my own here.

src/components/advanced-table/advanced-table.tsx
andrii-i: That's an interesting discussion. I believe both points of view are valid with concern that @dlqqq raised about @jweill-aws offer potentially solved by using `Rows X-Y` instead of just `X-Y`

src/components/advanced-table/advanced-table.tsx
```diff
@@ -153,8 +157,35 @@ export function AdvancedTable<
...
+        const loadedRows = rows?.length ?? 0;
+        if (onLastPage) {
+          // for some reason `to` is set incorrectly on the last page in
+          // server-side pagination, so we need to build the string differently
+          // in this case.
+          return trans.__('%1–%2 of %3', from, loadedRows, loadedRows);
+        } else {
+          return (
+            trans.__('%1–%2 of %3', from, to, loadedRows) +
+            (nextToken === undefined ? '' : '+')
```
jweill-aws, line 179: Localized strings shouldn't be concatenated with other strings, even for one character. This covers languages like Japanese where the `loadedRows` parameter wouldn't appear last.

```suggestion
            trans.__('%1–%2 of %3', from, to,
              loadedRows + (nextToken === undefined ? '' : '+'))
```

src/components/advanced-table/advanced-table.tsx
dlqqq: Let's discuss this as a team later, and we can open a follow-up PR if necessary.

jweill-aws (approved):

PR #206, opened 2022-10-26
Creates directory based on job name, not based on timestamp

### LINKED ISSUE #201: Directory name conflict if two jobs have the same input filename
## Description

If two jobs have the same input filename and execute within one second of each other, they will attempt to create the same directory under `jobs/`. This is destructive and will cause one of the jobs to fail.

## Reproduce

1. Create job A on filename F to run every hour on the hour.
2. Create job B on filename F to run every hour on the hour.
3. Wait for the top of the hour.
4. Only A or B runs successfully.

## Expected behavior

Both A and B run successfully.

## Context

We need to name subdirectories in a way that avoids conflict and without too much verbosity. (Appending the job ID would avoid conflict, but job IDs are long.)

jweill-aws:
Fixes #201.

Old file structure, using the filename "stem" and the run timestamp:

`jobs/plot-2022-10-26-12-0813-PM/plot.ipynb`

New file structure, using the filename stem and the job ID:

`jobs/plot-6d56ab09-da11-4599-a503-7c3b43c7e3e1/plot.ipynb`

This change also introduces a `-` delimiter between minutes and seconds above.

### DIFF
--- a/jupyter_scheduler/orm.py
+++ b/jupyter_scheduler/orm.py
@@ -8,7 +8,7 @@ from sqlalchemy import Boolean, Column, Integer, String, create_engine
 from sqlalchemy.orm import declarative_base, declarative_mixin, registry, sessionmaker
 
 from jupyter_scheduler.models import EmailNotifications, Status
-from jupyter_scheduler.utils import create_output_filename, get_utc_timestamp
+from jupyter_scheduler.utils import get_utc_timestamp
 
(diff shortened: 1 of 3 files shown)


andrii-i (approved):
This looks good to me. @dlqqq also wanted to review this PR so he might have more comments

3coins (approved):

PR #207, opened 2022-10-26
merge `ListJobsView` into `JobsView`, re-define `IAdvancedOptionsProps`

### LINKED ISSUE #105: Review front end models

### LINKED ISSUE #205: Model type is not correctly defined used in `handleModelChange` function
## Description
for the type in model, we are using `ICreateJobModel | IJobDetailModel | IJobDefinitionModel`
but in the `handleModelChange` method we are only using `ICreateJobModel`: https://github.com/jupyter-server/jupyter-scheduler/blob/main/src/tokens.ts#L17

This causes a TSC issue as the types are not set properly:
```
is missing the following properties from type 'ICreateJobModel': createType, scheduleInterval
```
We can bypass by marking model as `any` type, or disabling tsc for that line, but this is still an issue.

## Expected behavior
the model param for the function should be defined the same way as `model`

jweill-aws:
## Description

- Merges `ListJobsView` type into `JobsView` type
  - Previously, the variables controlling "which page to render" was split up into three places. Let `model` refer to an instance of `JobsModel`.
  - `model.jobsView`, `model.listJobsModel.listJobsView`, and `model.jobDetailModel.detailType`
- Makes `JobsView` an enum
- Exports `JobsView` enum from package index. i.e. `import { JobsView } from '@jupyterlab/scheduler'`.
- Refactors `IAdvancedOptionsProps` to be a union of types discriminated by the `jobsView` property, which provides the following guarantees:
  - developers writing advanced options form extensions can always narrow the type of `model` to the correct union member by discriminating unions via a simple `if` statement.
    ```ts
    if (props.jobsView === JobsView.CreateForm) {
      // model is guaranteed to be of type `CreateJobModel` here.
      ...
    }
    ```
  - `model` will always have the same type as the argument accepted by `handleModelChange()`, and thus avoids the need for ugly type casting when calling `props.handleModelChange()`.

## Issues closed

- Fixes #205 
- Fixes #105

### DIFF
--- a/src/advanced-options.tsx
+++ b/src/advanced-options.tsx
@@ -5,7 +5,7 @@ import { FormLabel, Stack, TextField } from '@mui/material';
 import { Cluster } from './components/cluster';
 import { AddButton, DeleteButton } from './components/icon-buttons';
 import { useTranslator } from './hooks';
-import { ICreateJobModel, IJobDetailModel } from './model';
+import { JobsView } from './model';
 import { Scheduler } from './tokens';
 
 const AdvancedOptions = (
@@ -15,16 +15,20 @@ const AdvancedOptions = (
 
   const trans = useTranslator('jupyterlab');
 
-  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) =>
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    if (props.jobsView !== JobsView.CreateForm) {
+      return;
+    }
+
     props.handleModelChange({
-      // Only the create-job model can change, not the detail model
-      ...(props.model as ICreateJobModel),
+      ...props.model,
       [e.target.name]: e.target.value
     });
+  };
 
   const handleTagChange = (event: ChangeEvent<HTMLInputElement>) => {
-    if (props.jobsView !== 'CreateJob') {
-      return; // Read-only mode
+    if (props.jobsView !== JobsView.CreateForm) {
+      return;
     }
 
     const { name, value } = event.target;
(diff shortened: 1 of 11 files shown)


dlqqq:
Pushed my changes, edited PR description and title.

src/advanced-options.tsx
```diff
@@ -15,16 +15,20 @@ const AdvancedOptions = (
 
   const trans = useTranslator('jupyterlab');
 
-  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) =>
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    if (props.jobsView !== JobsView.CreateForm) {
```
jweill-aws, line 19: Could be addressed later: `jobsView` and `JobsView` should be renamed, as they might refer to a singular job or one or more job definitions

src/mainviews/create-job.tsx
```diff
@@ -824,9 +826,9 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
             </AccordionSummary>
             <AccordionDetails id={`${formPrefix}create-panel-content`}>
               <props.advancedOptions
-                jobsView={'CreateJob'}
+                jobsView={JobsView.CreateForm}
                 model={props.model}
-                handleModelChange={props.handleModelChange}
+                handleModelChange={model => props.handleModelChange(model)}
```
jweill-aws, line 831: Are you sure this is still necessary?

dlqqq: Good catch. Reverted this change.

src/advanced-options.tsx
dlqqq: Let's track this in a separate issue. https://github.com/jupyter-server/jupyter-scheduler/issues/210

andrii-i (approved):

dlqqq (approved):
Approving on Jason's behalf.

PR #212, opened 2022-10-27
fix view for empty list

### LINKED ISSUE #211: Switching to empty list causes blank widget
## Description

Switching to an empty job definitions or jobs list causes the screen to go blank.

## Reproduce

1. Delete all jobs and job definitions.
2. Create a new job, bringing you to the job list
3. Click "Job Definitions List"
4. Widget goes blank, error displayed in console: `Rendered fewer hooks than expected. This may be caused by an accidental early return statement.`

## Expected behavior

Empty Job Definitions List appears.

## Context

Appears to be browser-independent

dlqqq:
Fixes #211.

Caused by an early return in case of an empty list, which caused the `labelDisplayedRows()` hook to be called conditionally, causing a runtime error.

### DIFF
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState, useMemo } from 'react';
 
 import { useTheme } from '@mui/material/styles';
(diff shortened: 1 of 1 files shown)


andrii-i (approved):
Tested locally, fixes empty list problem: I can see both empty jobs list and empty job definitions list now instead of an empty screen

PR #226, opened 2022-10-29
Use LabeledValue in advanced options in Job Definition and Job Definition Detail

### LINKED ISSUE #165: Editable/read-only state of Advanced Options in Job Definition Detail

andrii-i:
## Description 

Render `Idempotency Token` and `Tags` fields of `advancedOptions` component as `LabeledValue` instead of `TextField` if `advancedOptions` is displayed in Detail View. Version of PR #224 without addition of `readonly` prop to `IAdvancedOptionsSharedProps` interface.

Fixes #165 

## Preview
readonly_adbanced_options

### DIFF
--- a/src/advanced-options.tsx
+++ b/src/advanced-options.tsx
@@ -7,6 +7,7 @@ import { AddButton, DeleteButton } from './components/icon-buttons';
 import { useTranslator } from './hooks';
 import { JobsView } from './model';
 import { Scheduler } from './tokens';
+import { LabeledValue } from './components/labeled-value';
 
 const AdvancedOptions = (
   props: Scheduler.IAdvancedOptionsProps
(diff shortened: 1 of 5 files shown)


ellisonbg (approved):
Overall this is a solid improvement!

PR #227, opened 2022-10-29
Create job from job definition - UI

### LINKED ISSUE #209: Run now from a job definition
### Problem

Previously, we offered a way for users to "re-run" a previously run job, but this didn't leverage a job definition. We removed that feature as it was incompatible with the API change we were making (removing the `input_url`). However, we believe that users we will still want to rerun a job. 

### Proposed Solution

We proposed allowing users to "re-run" a job by adding a "Run now" action for a job definition. To implement this we would create a new REST API and `Scheduler` method and then build the UI on top of this. The reason we are thinking of using the job definition for this is as follows:

* The job definition already has an immutable snapshot of the input file that can be used for the job.
* The job definition allows us to show the user all of the jobs that are "grouped" by the job definition.

### LINKED ISSUE #225: "Create Job" form title is not localized
The title of the "Create Job" form is `Create Job`, not `trans.__('Create Job')`.

https://github.com/jupyter-server/jupyter-scheduler/blob/03cb4e9b8f05f57618860d5621398157aa92092d/src/mainviews/create-job.tsx#L756

jweill-aws:
Fixes #209 and #225.

On the Job Definition Detail page, adds a "Run Job" button that brings the user to a minimal version of the Create Job form that displays only the job definition ID and lets the user edit parameters. The API to create a job this way was added in #228.

Demo video:

https://user-images.githubusercontent.com/93281816/198753293-50c90be5-7e93-4307-879a-0ccb52816195.mov

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -184,7 +184,7 @@ export class SchedulerService {
     return data as Scheduler.ICreateJobResponse;
   }
 
-  async createJobFromJobDefinition(
+  async createJobFromDefinition(
     definition_id: string,
     parameters: { [key: string]: any }
   ): Promise<Scheduler.ICreateJobResponse> {
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -186,7 +186,7 @@ export class SchedulerService {
 
   async createJobFromDefinition(
     definition_id: string,
-    parameters: { [key: string]: any }
+    model: Scheduler.ICreateJobFromDefinition
   ): Promise<Scheduler.ICreateJobResponse> {
     let data;
     try {
@@ -195,7 +195,7 @@ export class SchedulerService {
         `job_definitions/${definition_id}/jobs`,
         {
           method: 'POST',
-          body: JSON.stringify(parameters)
+          body: JSON.stringify(model)
         }
       );
     } catch (e: any) {
@@ -407,6 +407,10 @@ export namespace Scheduler {
     output_filename_template?: string;
     output_formats?: string[];
     compute_type?: string;
+  }
+
+  export interface ICreateJobFromDefinition {
+    parameters?: Parameters;
   }
 
   export type Status =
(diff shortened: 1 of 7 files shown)


ellisonbg:
Overall, looks good. Wondering if we might want to name the action "Run job now" to emphasize that this is different from running it on a schedule. Also, the API in #228 will only allow users to set parameters, so we should hide the Additional options in this form. Also, will this form have the same job creation error handling as the normal create job (show the spinner, hide the buttons, show the error banner)?

jweill-aws:
I hesitated to use the word “now” because the button doesn’t take an immediate action, but rather brings the user to a form. (Perhaps “Run Job…” or “Run Job Now…” would be better.)

The button hiding behavior and error handling is the same as in the ordinary create job/JD form.

Sent from my iPhone

On Oct 29, 2022, at 12:12 PM, Brian E. Granger ***@***.***> wrote:

﻿

Overall, looks good. Wondering if we might want to name the action "Run job now" to emphasize that this is different from running it on a schedule. Also, the API in #228<https://github.com/jupyter-server/jupyter-scheduler/pull/228> will only allow users to set parameters, so we should hide the Additional options in this form. Also, will this form have the same job creation error handling as the normal create job (show the spinner, hide the buttons, show the error banner)?

jweill-aws:
(Apologies for the botched non-rebase; I hadn't set up Git properly on my new computer yet)

Updates this change to remove advanced options from the run-from-JD form. Still WIP because no API is available yet.

image

jweill-aws:
This change is blocked until #228 is merged.

3coins:
Request payload seems incorrect here, so passed parameters are not assigned. The payload should match the `CreateJobFromDefinition` model.
https://github.com/jupyter-server/jupyter-scheduler/blob/9dd885c967c98ee9bceb8046c066d1b2e5411de5/jupyter_scheduler/models.py#L272-L273

3coins (changes requested):

3coins (approved):

PR #239, opened 2022-10-31
fix pagination in AdvancedTable

dlqqq:
Pagination was broken by #212, as the list of rendered rows would not be updated after changing the page.


andrii-i (approved):
Fixes the bug

PR #264, opened 2022-11-02
refactor schedule inputs

### LINKED ISSUE #251: refactor ScheduleInputs component
## Description

When the schedule starts as `0 0 15 * *` (00:00 on the 15th of the month):

image

… and then the user switches to "Month" scheduling:

image

… the day of the month does not propagate to the input, but the time does.

## Reproduce

1. Schedule is 0 0 15 * *  (00:00 on the 15th of the month)
2. Click "Edit job definition"
3. Drop down "Interval," pick "Month"
4. Error: "Day of the month" is blank

## Expected behavior

Day of the month should be "15".

## Context

Found while validating #238.

### LINKED ISSUE #262: Job definition schedule errors are not reinstated when switching from JD to job and back
## Description

Job definition schedule errors do not reappear when the job definition (JD) schedule has an error, then the user switches to create a job, then they switch back to create a JD.

https://user-images.githubusercontent.com/93281816/199368871-26bb6d55-baed-4241-9465-c4f540ef6202.mov

## Reproduce

1. Open the create form for a file.
2. Click "Run on a schedule"
3. Change the time to `1` (invalid). Error appears in the field and "Create" is disabled.
4. Click "Run now". "Create" is enabled (#261).
5. Click "Run on a schedule". `1` reappears as the time, but without an error, and "Create" is enabled.

## Expected behavior

`1` reappears as the time. An error is shown on the Time field and "Create" is disabled.

## Context

The backend will still reject this job, but we should expect this to fail and disable "Create" in anticipation of this.

dlqqq:
## Description

- Refactors `ScheduleInputs` component and the corresponding UI model to remove duplicate state
- Removes duplicate logic in handling state changes
- Implements more robust **coercion logic**, i.e. populating inputs when switching to/from easy scheduling
- Automatically validates on mount and removes validation errors on unmount
- triggers validation on change

## Demo (outdated)

https://user-images.githubusercontent.com/44106031/199591081-e480e1f0-bc61-411d-9b61-6025d9976f01.mov

## Limitations

- Coercion logic has been kept as simple as possible to avoid maintainability issues. It parses each of the CRON terms via `parseInt()`, and uses a default only if the value cannot be parsed.
  - This means that inputting an invalid CRON expression like `99 99 * * MON-FRI`, then switching to "weekday" results in the time being set to `99:99`.
  - In other words, coercion generates an invalid output for invalid inputs.
  - I personally don't see this as a common edge case worth handling, as it requires the user to immediately input an invalid value and then switch intervals.
  - Furthermore, validation errors will prevent submission of an invalid coerced value

## Related issues

- Closes #251
- Closes #262

### DIFF
--- a/src/components/schedule-inputs.tsx
+++ b/src/components/schedule-inputs.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, useMemo } from 'react';
+import React, { useEffect, useMemo } from 'react';
 import { TranslationBundle } from '@jupyterlab/translation';
 
 import cronstrue from 'cronstrue';
@@ -7,6 +7,7 @@ import tzdata from 'tzdata';
 import {
   Autocomplete,
   FormControl,
+  FormHelperText,
   InputLabel,
   MenuItem,
   Select,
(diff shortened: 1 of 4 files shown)


jweill-aws:
Feedback from UX testing:

The "minutes of hour" interfere with the "time of day". If I select Hour, then pick 55, then switch to Day, the time is 00:55, but I had expected 00:00. Notably, this isn't bidirectional; if I change the time of day to 01:20, then switch back to Hour, the "minutes of hour" are 55, not 20. I think it would be better to keep "minutes of hour" and "time of day" (as hour and minute) as two separate state items.

Minor: 24-hour time is properly `hh:mm`, so 1:20 am is 01:20, but when I enter 01:20 and switch to a different interval (e.g., day to weekday, or vice versa) I see 1:20.

If I enter 33 as the day of the month, before I blur the control, I see helper text that the job will not run in months with fewer than 33 days:

image

When I blur the control, I see the expected 1–31 value validation message even after I refocus the control.

image


It might be better to impose errors on change, not just on blur, to prevent this.

Similarly, entering an invalid time causes a confusing helper text message, instead of an error:

image

dlqqq:
> The "minutes of hour" interfere with the "time of day". If I select Hour, then pick 55, then switch to Day, the time is 00:55, but I had expected 00:00. Notably, this isn't bidirectional; if I change the time of day to 01:20, then switch back to Hour, the "minutes of hour" are 55, not 20. I think it would be better to keep "minutes of hour" and "time of day" (as hour and minute) as two separate state items.

@jweill-aws They already are (`model.scheduleMinute` v.s. `model.scheduleClock`). I'm not able to reproduce this issue locally either:

https://user-images.githubusercontent.com/44106031/199602870-6e99a341-7800-45da-9c01-9bffeadf610c.mov

jweill-aws:
I also don't see the "minute of hour" propagating to "time of day" bug anymore after checking out your code and rebuilding. Must have been a problem on my end.

dlqqq:
Discussed this with @brgrange, we think it is better to validate on change rather than on blur. Just pushed this change.

jweill-aws:
00:00 is displayed in 12-hour time in the helper text as "0:00 AM" instead of "12:00 AM"

image

jweill-aws:
I created a job description to run at 2 minutes past the hour. When I edit it, I see "2" when I choose "Hour," and when I choose "Day", "Weekday", or "Week", the time is set to 00:02, which I hadn't expected.

https://user-images.githubusercontent.com/93281816/199615185-868d758c-9d9d-4cc4-a2cd-b64b01bf5e49.mov

dlqqq:
> 00:00 is displayed in 12-hour time in the helper text as "0:00 AM" instead of "12:00 AM"

@jweill-aws Fixed.

> I created a job description to run at 2 minutes past the hour. When I edit it, I see "2" when I choose "Hour," and when I choose "Day", "Weekday", or "Week", the time is set to 00:02, which I hadn't expected.

@jweill-aws I don't consider this a bug; it's more of an implementation detail of how coercion works. The user will have to edit the clock field if they go from a schedule running on an hourly basis to one running on a daily/weekly basis. We make the most logical choice (preserving as many cron terms as we can), but obviously this isn't going to cover cases in which coercion isn't even well-defined.

src/components/schedule-inputs.tsx
```diff
@@ -499,6 +385,81 @@ function parseTime(input: string): [number | undefined, number | undefined] {
...
+    ':' +
+    (minutes < 10 ? '0' + minutes : minutes)
+  );
+}
+
+type CronTerms = [number, number, number, number, number];
+
+/**
+ * Extracts cron terms and coerces them into an array of numbers. Ranges are
+ * coerced by their first term, e.g. "12-34" is coerced to 12.
```
jweill-aws, line 401: Technically, days of the week like `MON` (1) and `FRI` (5) are also valid cron terms, but this function would coerce them to 0 (Sunday).

jweill-aws (approved):

src/components/schedule-inputs.tsx
dlqqq: Agreed, this is expected of the current implementation.

PR #265, opened 2022-11-02
Adds UTC-only option to environment model

### LINKED ISSUE #255: Support UTC-based schedules as an alternate to timezones
### Problem

Some backends cannot support timezones and need to support UTC schedules instead.

### Proposed Solution

* An environment should have a new attribute that specifies if UTC or timezones are supported.
* When and env uses UTC, we will hide the timezone selector, and communicate to the user that all times are UTC.
* In that case, the timezone field will be empty in the create job request.

jweill-aws:
Fixes #255. Allows an environment to specify that it is UTC-only. A UTC-only environment will not have a time zone selector in the schedule section when creating job descriptions. In addition, when the user submits a job for such an environment, the `timezone` argument will not be sent in the request payload.

In the screen shot below, a custom handler has overridden an environment to be UTC-only.

Screen shot of create-job form showing UTC-only messaging

### DIFF
--- a/jupyter_scheduler/models.py
+++ b/jupyter_scheduler/models.py
@@ -25,6 +25,7 @@ class RuntimeEnvironment(BaseModel):
     metadata: Optional[Dict[str, str]]  # Optional metadata
     compute_types: Optional[List[str]]
     default_compute_type: Optional[str]  # Should be a member of the compute_types list
+    utc_only: Optional[bool]
 
     def __str__(self):
         return self.json()
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -519,7 +519,7 @@ export namespace Scheduler {
     metadata: { [key: string]: string };
     compute_types: string[] | null;
     default_compute_type: string | null;
-    utc_only: boolean | null;
+    utc_only?: boolean;
   }
 
   export interface IOutputFormat {
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -519,6 +519,7 @@ export namespace Scheduler {
     metadata: { [key: string]: string };
     compute_types: string[] | null;
     default_compute_type: string | null;
+    utc_only: boolean | null;
   }
 
   export interface IOutputFormat {
--- a/src/mainviews/create-job.tsx
+++ b/src/mainviews/create-job.tsx
@@ -113,11 +113,18 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
           envList[0].name
         )?.map(format => format.name);
 
+        // Does the default environment support time zones?
+        let newTimeZone = 'UTC';
+        if (!envList[0].utc_only) {
+          newTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+        }
+
         props.handleModelChange({
           ...props.model,
           environment: envList[0].name,
           computeType: newComputeType,
-          outputFormats: outputFormats
+          outputFormats: outputFormats,
+          timezone: newTimeZone
         });
       }
     };
@@ -160,6 +167,7 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
       const envObj = environmentList.find(env => env.name === target.value);
       // Validate that the default compute type is in fact in the list
       let newComputeType = envObj?.compute_types?.[0];
+
       if (
         envObj?.default_compute_type &&
         envObj?.compute_types &&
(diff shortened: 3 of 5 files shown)


ellisonbg (approved):
Approving, but would be good for @pijain to look through this to ensure everything is good on the server side.

src/handler.ts
```diff
@@ -519,6 +519,7 @@ export namespace Scheduler {
     metadata: { [key: string]: string };
     compute_types: string[] | null;
     default_compute_type: string | null;
+    utc_only: boolean | null;
```
dlqqq, line 522: I don't think `null` is a possible value for the backend response, since there's no `null` equivalent in Python. You should be free to remove this and the weird `?? undefined;` clause in `create-job.tsx`.

src/mainviews/create-job.tsx
```diff
@@ -160,22 +167,32 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
       const newEnvOutputFormats = outputFormatsForEnvironment(
         environmentList,
         target.value
       )?.map(format => format.name);
+
+      // Does the new environment support time zones?
+      let newTimeZone = 'UTC';
+      if (!envObj?.utc_only) {
+        newTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+      }
```
dlqqq, line 188: Prefer ternary over conditional assignment.

```ts
const envObj?.utc_only
  ? 'UTC'
  : Intl.DateTimeFormat().resolvedOptions().timeZone;
```

dlqqq (changes requested):
This looks great, aside from a few minor changes. We'll also need to pass the `utcOnly` prop in the `edit-job-definition.tsx` component.

jupyter_scheduler/models.py
```diff
@@ -25,6 +25,7 @@ class RuntimeEnvironment(BaseModel):
     metadata: Optional[Dict[str, str]]  # Optional metadata
     compute_types: Optional[List[str]]
     default_compute_type: Optional[str]  # Should be a member of the compute_types list
+    utc_only: Optional[bool]
```
3coins, line 28: The naming here seems a bit odd to me. Should we rename to `timezones` or something related and have it True for OSS?

src/mainviews/create-job.tsx
```diff
@@ -113,11 +113,18 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
           envList[0].name
         )?.map(format => format.name);
 
+        // Does the default environment support time zones?
+        let newTimeZone = 'UTC';
+        if (!envList[0].utc_only) {
+          newTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+        }
```
dlqqq, line 120: Ah, something else I noticed: this effect hook only gets called on initial render, since this is being used to fetch the initial environment list.

This means switching environments may not update the new timezone as expected.

dlqqq (changes requested):

jupyter_scheduler/models.py
jweill-aws: I thought about using `timezones`, but I rejected it, thinking that the user would expect `timezones` to be set to a list of time zones. We could say `supports_timezones` if we want this to be a positive, although we'd have to interpret "unset" as "true" to avoid breaking existing envs.

src/mainviews/create-job.tsx
```diff
@@ -160,22 +167,32 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+        newTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+      }
+
       props.handleModelChange({
         ...props.model,
         environment: target.value,
         computeType: newComputeType,
-        outputFormats: newEnvOutputFormats
+        outputFormats: newEnvOutputFormats,
+        timezone: newTimeZone
```
dlqqq, line 195: The problem with doing this on select is that it resets the timezone even if you're switching between environments that support timezones. For example: browser's current timezone is 'X'. User sets to 'Y'. User then changes environments, and the timezone is always reset to 'X' even when it doesn't need to be.

dlqqq (changes requested):

jupyter_scheduler/models.py
dlqqq: I support preserving backwards compatibility by default, especially now that we're on 1.0.

dlqqq (approved):

PR #277, opened 2022-11-04
Adds error reporting

### LINKED ISSUE #236: Graceful handling for create form errors that cause a blank screen
### Problem

In cases of a fatal error, the create form disappears entirely, forcing a white or black screen, depending on theme:

https://user-images.githubusercontent.com/93281816/199113334-ed5e2a9d-0653-477a-a63b-538faf45aabc.mov

### Proposed Solution

Either display a message inside the widget with the captured error, or redirect users to the list view, in this case.

jweill-aws:
Fixes #236. Adds an error handler that intercepts errors and displays a message.

In the example below, the create form has been overridden to throw an exception.

open-create-error

### DIFF
--- a/src/components/error-boundary.tsx
+++ b/src/components/error-boundary.tsx
@@ -1,0 +1,66 @@
+import { Alert, AlertTitle, Stack, Typography } from '@mui/material';
+
+import React from 'react';
+
+interface IErrorBoundaryProps {}
+
+interface IErrorBoundaryState {
+  hasError: boolean;
+  error?: unknown;
+}
+
+export class ErrorBoundary extends React.Component<
+  React.PropsWithChildren<IErrorBoundaryProps>,
+  IErrorBoundaryState
+> {
+  constructor(props: IErrorBoundaryProps) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(error: unknown): IErrorBoundaryState {
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true, error: error };
+  }
+
+  componentDidCatch(error: unknown, errorInfo: React.ErrorInfo): void {
+    // You can also log the error to an error reporting service
+    console.log('Caught error in ErrorBoundary', error, errorInfo);
+  }
+
+  render(): React.ReactNode {
+    const errorMessage =
+      'We encountered an internal error. Please try your command again.';
+
+    let errorDetail;
+    if (typeof this.state.error === 'string') {
+      errorDetail = this.state.error;
+    } else if (this.state.error instanceof Error) {
+      errorDetail = this.state.error.message;
+    }
+
+    let infoSection;
+    if (errorDetail !== undefined) {
+      infoSection = (
+        <>
+          <Typography variant="h1">Error details</Typography>
+          <pre>{errorDetail}</pre>
+        </>
+      );
+    }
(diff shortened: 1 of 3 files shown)


src/notebook-jobs-panel.tsx
```diff
@@ -90,68 +93,82 @@ export class NotebookJobsPanel extends VDomRenderer<JobsModel> {
       this.model.jobsView = JobsView.CreateForm;
     };
 
+    const trans = this._translator.load('jupyterlab');
+
```
dlqqq, line 97: This can be defined in `constructor()` and bound to an instance variable instead of loading this every render.

src/components/error-boundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+> {
+  constructor(props: IErrorBoundaryProps) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError(error: unknown): IErrorBoundaryState {
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true, error: error };
+  }
```
dlqqq, line 28: Doesn't look like this is being used.

src/components/error-boundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+    this.setState({ hasError: true, error });
+  }
+
+  render(): React.ReactNode {
+    let errorDetail;
+    if (typeof this.state.error === 'string') {
+      errorDetail = this.state.error;
+    } else if (this.state.error instanceof Error) {
+      errorDetail = this.state.error.message;
+    }
```
dlqqq, line 42: The problem here is that if the raised error is not an Error object or a string, then the error detail doesn't get shown. Since all objects contain a `toString()` method in JS, you should be able to do:

```tsx
const errorDetail = this.state.error instanceof Error
  ? this.state.error.message
  : this.state.error.toString();
```

src/components/error-boundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+    if (this.state.hasError) {
+      return (
+        <div className="jp-error-boundary">
+          <Stack spacing={4}>
+            <Alert severity="error">
+              <AlertTitle>{this.props.alertTitle}</AlertTitle>
+              {this.props.alertMessage}
+            </Alert>
+            {infoSection}
```
dlqqq, line 61: Nit: Avoid assigning JSX elements to temporary variables, since it's not clear what the variable is for when reading the code top-down (i.e. it requires two passes).

I would prefer:

```
{errorDetail && (
  <> ... </>
)}
```

dlqqq (changes requested):

src/components/error-boundary.tsx
dlqqq: Obviously, seeing `[Object object]` would not be particularly helpful, but it would be better than nothing.

jweill-aws: Unfortunately, `unknown.toString()` doesn't pass `tsc`

dlqqq: Hmm, looks like `toString()` can't be called on `null` or `undefined`. I think the safest case is to keep the current implementation rather than risk triggering a second exception and rendering a blank screen when catching one.

dlqqq (approved):

PR #280, opened 2022-11-04
hide timezone selector when editing job definition created with utc_only environment

### LINKED ISSUE #279: Timezone selector appears when editing job definitions in a UTC-only environment
## Description

When editing a job definition that runs in an environment that only accepts schedules in UTC, the time zone selector appears in the edit form.

Create job form showing the time zone picker removed

Edit job description form showing time zone picker

## Reproduce

1. Create a job definition (JD) in an environment with `utc_only=true`
2. Edit the JD
3. **Bug**: Time zone picker appears, set to UTC.

## Expected behavior

The time zone picker should be replaced with explanatory UTC text, as in the screen shot above.

## Context

Changes introduced in #265.

dlqqq:
Fixes #279.

New behavior of edit job definition view:

- fetches list of environments on initial render
- sets model timezone and passes `utcOnly: true` prop to `ScheduleInputs` component accordingly

Also updates the seed script to use the conda environments in the job definitions.

https://user-images.githubusercontent.com/44106031/200087959-b0f31558-1998-46d0-9c08-dabf0a010969.mov

### DIFF
--- a/dev/seed.py
+++ b/dev/seed.py
@@ -36,7 +36,7 @@ def get_db_path():
     return os.path.join(data[0], "scheduler.sqlite")
 
 
-def create_random_job_def(index: int) -> CreateJobDefinition:
+def create_random_job_def(index: int, env: str) -> CreateJobDefinition:
     name = (
         random.choice(
             [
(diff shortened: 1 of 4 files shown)


jweill-aws (approved):
Tested locally successfully

andrii-i (commented):
Looks and works well 👍

PR #284, opened 2022-11-08
Error alerts in list job definitions view

### LINKED ISSUE #276: Show errors from Resume action
### Problem

When the user clicks "Resume" on a paused Job Definition, and the resume action fails, they do not see an error.

### Proposed Solution

On the list and details views, when the resume request returns an error, display the error in the UI.

- [x] List view (#284)
- [ ] Detail view (#286)

jweill-aws:
Fixes #276 by capturing failures of pause, reject, and delete job definition actions and displaying them as alerts.

This only affects the job definitions list. The jobs list's "delete" action is asynchronous; there are no other immediate actions.

reject-pause

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -60,7 +60,7 @@ export class SchedulerService {
           method: 'GET'
         }
       );
-    } catch (e: any) {
+    } catch (e) {
       console.error(e);
     }
     return data as Scheduler.IDescribeJobDefinition;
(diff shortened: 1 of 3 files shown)


src/handler.ts
```diff
@@ -113,7 +113,7 @@ export class SchedulerService {
         }
       );
     } catch (e: any) {
```
andrii-i, line 115: No need to explicitly type error to `any` or `unknown` anywhere in this file

andrii-i (approved):
Looks and works well. 2 minor style comments for your consideration

PR #286, opened 2022-11-08
Errors for pause, resume, delete in detail page and list view

### LINKED ISSUE #276: Show errors from Resume action

jweill-aws:
Fixes #276. Adds error handlers for the detail page and list view for pause, resume, and delete for jobs and job definitions (except for "delete job" in the list view, which does not make a command call). Displays an alert if an error gets caught. Suppresses the alert on a reload or a new command.

In the video below, the handler is customized to throw errors on all pause and delete job definition requests.

https://user-images.githubusercontent.com/93281816/200671964-4eda0ba6-80f8-4cba-9c76-8001dd066d8b.mov

### DIFF
--- a/src/components/job-definition-row.tsx
+++ b/src/components/job-definition-row.tsx
@@ -92,7 +92,7 @@ export function buildJobDefinitionRow(
   forceReload: () => void,
   trans: TranslationBundle,
   ss: SchedulerService,
-  handleApiError: (error: string | undefined) => void
+  handleApiError: (error: string | null) => void
 ): JSX.Element {
   const cellContents: React.ReactNode[] = [
     // name
@@ -110,7 +110,7 @@ export function buildJobDefinitionRow(
       <PauseButton
         jobDef={jobDef}
         clickHandler={async () => {
-          handleApiError(undefined);
+          handleApiError(null);
           ss.pauseJobDefinition(jobDef.job_definition_id)
             .then(_ => {
               forceReload();
@@ -123,7 +123,7 @@ export function buildJobDefinitionRow(
       <ResumeButton
         jobDef={jobDef}
         clickHandler={async () => {
-          handleApiError(undefined);
+          handleApiError(null);
           ss.resumeJobDefinition(jobDef.job_definition_id)
             .then(_ => {
               forceReload();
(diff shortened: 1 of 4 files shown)


src/components/job-definition-row.tsx
```diff
@@ -92,7 +92,7 @@ export function buildJobDefinitionRow(
   forceReload: () => void,
   trans: TranslationBundle,
   ss: SchedulerService,
-  handleApiError: (error: string) => void
+  handleApiError: (error: string | undefined) => void
```
andrii-i, line 95: `handleApiError` accepts `undefined` because it's one of the expected inputs / error object can be `undefined` right? 

(If undefined is used to indicate absence of value and function does not necessary expects to receive undefined otherwise it I think it is better to use `null` to indicate an absence of value s.t. there is a clear differentiation vs 
`undefined` and function would not accept just non-existing/unset field)

src/mainviews/detail-view/job-definition.tsx
```diff
@@ -33,22 +40,28 @@ export interface IJobDefinitionProps {
 
 export function JobDefinition(props: IJobDefinitionProps): JSX.Element {
   const trans = useTranslator('jupyterlab');
+  const [displayError, setDisplayError] = useState<string | undefined>();
```
andrii-i, line 43: Same as in https://github.com/jupyter-server/jupyter-scheduler/pull/286/files#r1017133144

src/mainviews/detail-view/job-detail.tsx
```diff
@@ -62,11 +62,12 @@ export const timestampLocalize = (time: number | ''): string => {
 export function JobDetail(props: IJobDetailProps): JSX.Element {
   const trans = useTranslator('jupyterlab');
   const [downloading, setDownloading] = useState(false);
+  const [displayError, setDisplayError] = useState<string | undefined>();
```
andrii-i, line 65: Same as [comment above](https://github.com/jupyter-server/jupyter-scheduler/pull/286/files#r1017133144)

src/mainviews/detail-view/job-detail.tsx
```diff
@@ -62,11 +62,12 @@ export const timestampLocalize = (time: number | ''): string => {
...
 
   const ss = new SchedulerService({});
 
   const translateStatus = useCallback(
-    (status: Scheduler.Status) => {
+    (status: Scheduler.Status | undefined) => {
```
andrii-i, line 70: Same as [comment above](https://github.com/jupyter-server/jupyter-scheduler/pull/286/files#r1017133144)

andrii-i (approved):

PR #287, opened 2022-11-08
Adds tooltips to job definition links from list view

### LINKED ISSUE #269: Tooltip for job name link
### Problem

In the list view, most links, as of #266, have tooltips, but the "Job name" link does not.

### Proposed Solution

Add a tooltip that says "View details for *(job name)*" when the user hovers over the link in the "Job name" column.

Add a tooltip that says "View details for *(job definition name)*" for the "Job definition name" column in the Notebook Job Definitions list view.

### Additional context

Providing titles for links improves accessibility.

jweill-aws:
Fixes #269. In the jobs list, job names already had tooltips on hover. Adds the same functionality, with the same styling, to the job definitions list.

image


andrii-i (approved):
Good catch!

PR #292, opened 2022-11-09
correctly handle last page even when latest next_token is truthy

### LINKED ISSUE #289: Table displays "6–5 of 5" or similar when there is a next page token with no additional records
## Description

When the final page of results includes a next-page token, but the next page has no results, the "from" number is larger than the total rows number.

image

## Reproduce

1. Go to the last page of results of a table populated from a backend that incorrectly sets the "next page" token on the last page of results.
2. Click the next page button.
3. Error: "6–5 of 5" or similar appears

## Expected behavior

An alternative message or no message displays.

### LINKED ISSUE #291: Next page error handling improvements
### Problem

This is a follow up to #289.

In a list view's Advanced Table, if the user clicks "next page" and there are no items on the next page (e.g., because of a job deletion after the page loaded), keep the user on the current page, potentially showing them an alert message.

### Proposed Solution

Keep the user on the current page and display an alert message.

(Should the "next page" button be deleted at this point, or should we keep it active and let users try again?)

dlqqq:
Renders error banner in `AdvancedTable` when attempting to navigate to a next page with no rows. We determine that the user is already on the last page and disable the next page button as well.

## Demos

https://user-images.githubusercontent.com/44106031/200930566-b732e602-a9d8-47ee-9d8e-97590e6c9c13.mov

https://user-images.githubusercontent.com/44106031/200930588-b40c7fee-0804-4535-a8fa-9bfed606efe2.mov

## Related issues / PRs

- Closes #289
- Reverts #290
- Closes #291

### DIFF
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -155,8 +155,6 @@ export function AdvancedTable<
       .catch((e: Error) => {
         setDisplayError(e.message);
       });
-
-    return true;
   };
 
   const renderedRows: JSX.Element[] = useMemo(
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -113,10 +113,10 @@ export function AdvancedTable<
     fetchInitialRows();
   }, [props.query]);
 
-  const fetchMoreRows = async () => {
+  const fetchMoreRows = async (newPage: number) => {
     // Do nothing if the next token is undefined (shouldn't happen, but required for type safety)
     if (nextToken === undefined) {
-      return;
+      return false;
     }
 
     // Apply the custom token to the existing query parameters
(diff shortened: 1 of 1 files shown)


src/components/advanced-table/advanced-table.tsx
```diff
@@ -131,14 +131,26 @@ export function AdvancedTable<
       })
       .then(payload => {
         setLoading(false);
+        const newRows = props.extractRows(payload) || [];
+
+        if (newRows.length === 0) {
+          // no rows in next page -- leave page unchanged, disable next page
+          // button, and show an error banner
+          setNextToken(undefined);
+          setDisplayError(trans.__('Last page reached.'));
```
jweill-aws, line 140:
```suggestion
          setDisplayError(trans.__('This is the last page.'));
```

This may seem contradictory (why display a "next" button otherwise?) but it might be less jarring to see.

jweill-aws (commented):
What kind of testing can be done on this?

src/components/advanced-table/advanced-table.tsx
jweill-aws: Also, could this be somehow less severe than "error"? A less severe [alert](https://mui.com/material-ui/react-alert/), even "Info" (with state set in `setDisplayInfo`) would be more calming.

dlqqq: 1. I'd prefer to keep the string as is; it's more concise and conveys the same message.
2. I consider this an error, given that this should not happen for a backend correctly implementing `next_token`.

dlqqq:
@jweill-aws You can test this by modifying the logic in `scheduler.py` and making sure the backend always returns a `next_token` even on the last page. Specifically, for list jobs, on line 463:

```py
list_jobs_response = ListJobsResponse(
    jobs=jobs_list,
    next_token=next_token or "asdf",
    total_count=total,
)
```

jweill-aws (approved):

PR #297, opened 2022-11-10
Display success message on list view after creating a job

### LINKED ISSUE #293: Show something in job/definition list on successful creation
### Problem

A user creates a job or a job definition. The service backend does not make that job or job definition visible immediately, so the newly-created item does not show up in the list immediately.

### Proposed Solution

After the user creates a job or a job definition, and the request has succeeded, bring the user to the list view with an entity (hereinafter the "fake item") added to the top as if the job or job definition ("real item") has been created. The fake item should have the metadata, including ID, that the real item will have after it is present in the backend. There should be a link to the job/definition detail page from the fake item's name.

If the query result includes a real item with an ID matching the just-created item, then do not display a fake item.

After the user clicks Reload, display the query result exactly as the backend sent it, without any fake item.

jweill-aws:
Fixes #293.

Adds a success message to the list jobs page after creating a new job.

After we create a job, display a "success" banner on the list page. The message reads, "Your job/definition 'Foo' has been created. If you do not see it in the list below, please reload the list in a few seconds."

The message passed to the list page will include an expiration time for the success message. If the page is loaded after that time, we will not display the success banner. This is to avoid awkwardness if the user reloads the page a long time later (e.g., due to a session restoration).

info-message

This change also fixes some minor localization bugs and makes both the "info" and "error" banners on list pages closable with an "X" button on each.

### DIFF
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -114,6 +114,14 @@ export function AdvancedTable<
         setNextToken(payload?.next_token);
         setTotalCount(payload?.total_count);
 
+        // Run find condition
+        if (
+          props.findCondition !== undefined &&
+          props.infoMessageIfNotFound !== undefined
+        ) {
+          console.log('Running find condition');
+        }
+
         // Check the initial rows
         if (
           props.findCondition !== undefined &&
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -77,10 +77,7 @@ export function AdvancedTable<
   const [page, setPage] = useState<number>(0);
   const [maxPage, setMaxPage] = useState<number>(0);
   const [loading, setLoading] = useState<boolean>(true);
-  const [displayError, setDisplayError] = useState<React.ReactNode | null>(
-    null
-  );
-  const [displayInfo, setDisplayInfo] = useState<React.ReactNode | null>(null);
+  const [displayError, setDisplayError] = useState<string | null>(null);
   const trans = useTranslator('jupyterlab');
   const theme = useTheme();
 
@@ -101,8 +98,7 @@ export function AdvancedTable<
       })
       .then(payload => {
         setLoading(false);
-        const rows = props.extractRows(payload);
-        setRows(rows);
+        setRows(props.extractRows(payload));
         setNextToken(payload?.next_token);
         setTotalCount(payload?.total_count);
       })
@@ -262,16 +258,7 @@ export function AdvancedTable<
 
   return (
     <>
-      {displayError && (
-        <Alert severity="error" onClose={() => setDisplayError(null)}>
-          {displayError}
-        </Alert>
-      )}
-      {displayInfo && (
-        <Alert severity="info" onClose={() => setDisplayInfo(null)}>
-          {displayInfo}
-        </Alert>
-      )}
+      {displayError && <Alert severity="error">{displayError}</Alert>}
       {tableDiv}
     </>
   );
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -60,6 +60,14 @@ type AdvancedTableProps<
    * manually set to whatever value is provided. Defaults to 'auto'.
    */
   height?: 'auto' | string | number;
+  /**
+   * Optional condition that searches the first set of rows for a match. If the
+   * search fails, we will display an info message. If a findCondition is
+   * specified, an infoMessageIfNotFound must be specified, or the findCondition
+   * will not be tested.
+   */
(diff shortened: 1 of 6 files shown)


andrii-i (approved):

src/components/advanced-table/advanced-table.tsx
```diff
@@ -77,7 +77,10 @@ export function AdvancedTable<
   const [page, setPage] = useState<number>(0);
   const [maxPage, setMaxPage] = useState<number>(0);
   const [loading, setLoading] = useState<boolean>(true);
-  const [displayError, setDisplayError] = useState<string | null>(null);
+  const [displayError, setDisplayError] = useState<React.ReactNode | null>(
+    null
+  );
+  const [displayInfo, setDisplayInfo] = useState<React.ReactNode | null>(null);
```
dlqqq, line 83: `setDisplayInfo()` is never called and thus this state variable is never used meaningfully. It looks like the only significant change is that the display error is dismissible. Could you revert all the other changes in this file?

src/mainviews/list-jobs.tsx
```diff
@@ -185,7 +220,29 @@ function ListJobDefinitionsTable(props: ListJobDefinitionsTableProps) {
...
+      ? trans.__(
+          'Your job definition "%1" has been created. ' +
+            'If you do not see it in the list below, please reload the list in a few seconds.',
+          props.newlyCreatedName
+        )
+      : null;
+
+  const [displayInfo, setDisplayInfo] = useState<React.ReactNode | null>(
+    successMessage
+  );
```
dlqqq, line 245: 1. This state management logic is faulty. It only clears the error on re-render, not every minute as expected.

2. The success message expiration time is passed as a prop even though it's a constant value. This makes it confusing because now you've split the expiration logic into `notebook-jobs-panel.tsx` for no reason. Instead, hardcode a constant like `const INFO_BANNER_EXPIRATION = 60`, expose it here in the list view, and remove `successMessageExpiration`.

3. You are duplicating logic in both the ListJobs and ListJobDefinitions table, when you could instead place this logic inside their shared parent component `<NotebookJobsList />`

In short, your component logic should look something like this:

```tsx
// outside component at file-scope
const INFO_BANNER_EXPIRATION = 60

// inside <NotebookJobsList /> component
const [displayInfo, setDisplayInfo] = useState<React.ReactNode>(null);
useEffect(() => {
  if (!props.newlyCreatedId && !props.newCreatedName) return;
  const successMessage = ...;
  setDisplayInfo(successMessage)
  setTimeout(() => setDisplayInfo(null), INFO_BANNER_EXPIRATION);
}, [props.newlyCreatedId, props.newlyCreatedName]);
```

dlqqq (changes requested):

src/mainviews/list-jobs.tsx
jweill-aws: The reason for passing the expiration time as a prop is so that it gets serialized, not so that the banner disappears after 60 seconds (which is not a requirement). This solves the problem of:

1. User creates a job
2. User is brought to the list jobs page which displays a success banner
3. User closes their browser
4. Hours later, the user reconnects to JupyterLab
5. User should see the list jobs page without a success banner (because they haven't done anything recently).

dlqqq: OK, then it sounds like the real solution is to not have the newly created ID and name serialized in the model, correct? IOW, just have them be instance properties on the NotebookJobsPanel class.

dlqqq: I don't see a use case for making our logic more complex to support the use case of a persistent banner across reloads within 60 seconds.

jweill-aws: Deduplicated status message. Removed the 60-second timeout.

src/notebook-jobs-panel.tsx
```diff
@@ -62,7 +62,13 @@ export class NotebookJobsPanel extends VDomRenderer<JobsModel> {
...
   }
 
-  showListView(view: JobsView.ListJobs | JobsView.ListJobDefinitions): void {
+  showListView(
+    view: JobsView.ListJobs | JobsView.ListJobDefinitions,
+    newlyCreatedId?: string,
+    newlyCreatedName?: string
+  ): void {
+    this.model.listJobsModel.newlyCreatedId = newlyCreatedId;
+    this.model.listJobsModel.newlyCreatedName = newlyCreatedName;
```
dlqqq, line 71: Wait, but binding these attributes to the list jobs model persists them. That means even when I refresh the page, don't these attributes remain bound and thus the banner persists?

I had meant for these attributes to be stored as plain old properties on the NotebookJobsPanel class (outside of `this.model`), which are not persisted.

dlqqq (changes requested):

dlqqq (approved):

Issue #296, opened 2022-11-09
Additional feature handlers for UI buttons


jweill-aws:
### Problem

Some buttons in the UI that execute a command or issue an API request do not handle errors resulting from that action. They either ignore the error, display a generic error boundary message, or do something else undesired.

### Proposed Solution

Intercept and display errors in the following contexts:

#### List notebook jobs

- [ ] Download files
- [ ] Delete
- [ ] Stop

#### Job detail

- [ ] Download files
- [x] Delete (fix in #302)

### Additional context

For examples of changes that added error handlers, see #284, #286, and #288.

The `handlers.ts` file should throw an error (`return Promise.reject(e)`) when all calls related to the above buttons fail.

PR #302, opened 2022-11-11
Added exception handling to both api handlers and UI

3coins:
- Added missing exception handling for API handlers, fixes catching and displaying these exceptions on the UI
- Added a new scheduler `SchedulerWithErrors` to aid UI testing of exceptions. This uses the default scheduler, but will randomly raise exceptions to simulate error conditions.

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -61,7 +61,7 @@ export class SchedulerService {
         }
       );
     } catch (e) {
-      return Promise.reject(e)
+      return Promise.reject(e);
     }
     return data as Scheduler.IDescribeJobDefinition;
   }
@@ -79,7 +79,7 @@ export class SchedulerService {
         method: 'GET'
       });
     } catch (e) {
-      return Promise.reject(e)
+      return Promise.reject(e);
     }
     return data as Scheduler.IListJobDefinitionsResponse;
   }
@@ -129,7 +129,7 @@ export class SchedulerService {
         method: 'GET'
       });
     } catch (e) {
-      return Promise.reject(e)
+      return Promise.reject(e);
     }
     return data as Scheduler.IDescribeJob;
   }
@@ -146,7 +146,7 @@ export class SchedulerService {
         method: 'GET'
       });
     } catch (e) {
-      return Promise.reject(e)
+      return Promise.reject(e);
     }
     return data as Scheduler.IListJobsResponse;
   }
@@ -163,7 +163,7 @@ export class SchedulerService {
         method: 'GET'
       });
     } catch (e) {
-      return Promise.reject(e)
+      return Promise.reject(e);
     }
 
     return data.count;
@@ -211,7 +211,7 @@ export class SchedulerService {
         body: JSON.stringify({ status })
       });
     } catch (e) {
-      return Promise.reject(e)
+      return Promise.reject(e);
     }
   }
 
@@ -222,7 +222,7 @@ export class SchedulerService {
         method: 'GET'
       });
     } catch (e) {
-      return Promise.reject(e)
+      return Promise.reject(e);
     }
 
     return data as Scheduler.IRuntimeEnvironment[];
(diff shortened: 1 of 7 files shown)


src/mainviews/detail-view/job-definition.tsx
```diff
@@ -231,10 +231,12 @@ export function JobDefinition(props: IJobDefinitionProps): JSX.Element {
   return (
     <>
       {displayError && <Alert severity="error">{displayError}</Alert>}
+      {props.model.inputFile && <>
```
jweill-aws, line 234: Run eslint on this block; it will complain

src/mainviews/detail-view/job-detail.tsx
```diff
@@ -304,10 +306,12 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
       {props.model.statusMessage && (
         <Alert severity="error">{props.model.statusMessage}</Alert>
       )}
-      {ButtonBar}
-      {CoreOptions}
-      {Parameters}
-      {AdvancedOptions}
+      {props.model.jobId && <>
```
jweill-aws, line 309: Ditto eslint

jweill-aws:
When I'm using `SchedulerWithErrors`, on the "edit job description" form, when I click "Save Changes" and an error occurs, no error banner appears on the edit form

jweill-aws:
When I'm on the "run job from job definition" screen and I create, and then the list jobs API fails, I see the list job error on the create-from-definition form. It's not clear from this message whether the new job was created.

image

3coins:
> When I'm on the "run job from job definition" screen and I create, and then the list jobs API fails, I see the list job error on the create-from-definition form. It's not clear from this message whether the new job was created.
> 
> image

Yes, the interaction is not going to be perfect since the failures are happening randomly. You can also go inside the Scheduler API and deliberately fail a specific method.

jweill-aws:
image

By repeatedly pausing and resuming a job definition, I can get the JD detail page into a state where the alert banner doesn't appear and nothing else does, either

jweill-aws (approved):

src/mainviews/detail-view/detail-view.tsx
```diff
@@ -63,17 +63,40 @@ export function DetailView(props: IDetailViewProps): JSX.Element {
...
+      setJobDefinitionModel({
+        definitionId: '',
+        name: '',
+        inputFile: '',
+        outputPath: '',
+        environment: '',
+        createError: e.message
+      });
+    }
   };
```
dlqqq, line 100: How about instead of setting this on the model, we keep the state local? The model should only be defined in the case where we successfully fetched a job / job definition from the backend, and we should keep it null otherwise. This would mean we need a separate state variable that is maintained as follows:

```tsx
const [fetchError, setFetchError] = useState<string>();

...

try {
  ...
} catch (e: Error) {
  setFetchError(e.message);
}
```

And right before rendering the body of the page, we display the error banner if necessary:

```tsx
return (
  ...
  {fetchError && <Alert severity="error">{fetchError}</Alert>}
  {props.jobsView === JobsView.JobDetail && jobModel && (
     ...
   )}
  ...
)
```

Finally, you should revert your changes to job definition and job detail.

dlqqq (changes requested):

src/handler.ts
```diff
@@ -270,7 +270,7 @@ export class SchedulerService {
         }
       );
     } catch (e) {
-      console.error(e);
+      console.log(e);
```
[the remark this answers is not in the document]
jweill-aws: This is about download files, which indeed doesn't have an error handler — see #296

src/mainviews/detail-view/detail-view.tsx
dlqqq: Just pushed a commit that implements my suggestions.

dlqqq (approved):

PR #315, opened 2022-11-17
Adds reload on detail page

### LINKED ISSUE #313: Add "reload" on Job Detail and Job Definition Detail if load failed
### Problem

If a job detail or job definition detail page fails to load because of a failed `GET` request, the page is mostly blank, and the only way out is to either close the panel or go back to the list:

image

### Proposed Solution

Add a "Reload" button on this page that attempts to load the job or job definition again.

jweill-aws:
Fixes #313. On top of the job detail and job definition detail page, adds a "Reload" button. This does not appear in the same button bar as other buttons do because those buttons only render if the job model or JD model is defined.

https://user-images.githubusercontent.com/93281816/202331296-95ebe62b-f56a-4bf9-96f3-a6f8935a8482.mov

### DIFF
--- a/src/mainviews/detail-view/job-definition.tsx
+++ b/src/mainviews/detail-view/job-definition.tsx
@@ -70,7 +70,7 @@ export function JobDefinition(props: IJobDefinitionProps): JSX.Element {
   const model: IJobDefinitionModel = props.model;
 
   const handleDeleteJobDefinition = async () => {
-    ss.deleteJobDefinition(props.model?.definitionId ?? '')
+    ss.deleteJobDefinition(model.definitionId ?? '')
       .then(_ => props.setJobsView(JobsView.ListJobDefinitions))
       .catch((e: Error) => setDisplayError(e.message));
   };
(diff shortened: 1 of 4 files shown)


andrii-i:
I would put all buttons in the same horizontal row and conditionally render which buttons are actually rendered. Red stop button would still be the main / rightmost button. It means that buttons would change their position depending on the state of the page which we tried to avoid previously in e.g. list view. Still I think it's preferable to having 2 rows of buttons

jweill-aws:
Updates to show "Reload Job" and "Reload Job Definition" buttons in a consistent button bar. (The more verbose names are to avoid a conflict with the "Reload" button on the list view in a JD detail page.)


https://user-images.githubusercontent.com/93281816/202580318-6977f3ee-1716-4acf-b390-f3014f2bfb91.mov

src/mainviews/detail-view/job-definition.tsx
```diff
@@ -20,62 +10,97 @@ import {
...
+
+  if (props.model === null) {
+    return (
+      <>
+        {ErrorBanner}
+        <ButtonBar>{ReloadButton}</ButtonBar>
+      </>
+    );
+  }
+  const model: IJobDefinitionModel = props.model;
```
andrii-i, line 70: On the lines after this, in some places this new `model` variable is used and in other places `props.model` is used. Is there a special reason to do this?

andrii-i (approved):
This is nice! I think this is more straightforward UX and cleaner UI. Thank you for optimizing `detail-view/*` code

PR #333, opened 2023-01-06
fix(event-note-icon): add viewBox attr to svg so it resizes

maxime-helen:
Bringing this fix as Sagemaker Studio couldn't resize the icon associated to "scheduling:show-notebook-jobs" command, added to the new Studio launcher with a smaller size than before

See this codepen for more infos https://codepen.io/tefa-dev/pen/gOjLZQZ, the icon doesn't resize without `viewBox`


dlqqq (approved):
@maxime-helen Great catch, thank you! CI failures not related.

PR #340, opened 2023-02-23
WIP: Adds error message on 404 response from backend

JasonWeill:
Work-in-progress change for #329.

If the server extension is absent or disabled, displays a dialog box after the user attempts to create a job or job definition:

image

… then, after the dialog box is dismissed, the user sees an informative message instead of a lot of HTML.

image (1)


dlqqq:
Superseded by #341.

PR #382, opened 2023-06-10
Fix two-phase delete buttons with Safari

JasonWeill:
Fixes #364 for Safari (desktop macOS) compatibility by removing the `onBlur` event on the `ConfirmButton` component. We had presumed that, after clicking the initial X icon, the focus would move to the `Button` component that did the actual delete, but in Safari, this is not the case.

As a noticeable UI change, this means that it's possible to have more than one red "Delete" button visible at a time. It is also no longer possible to cancel a two-phase Delete without reloading the view or the page, but I expect that this is a rare use case.

delete-buttons

Issue #364, opened 2023-03-31
Cannot delete jobs in Safari browser


ChaonengQuan:
## Description

I cannot delete notebook jobs in Safari browser. However, if I open jupyter lab in Firefox then notebook jobs can be deleted.

## Reproduce

1. Create a conda environment  `conda create -n py310 python=3.10 -y`
2. Activate the condo environment `conda activate py310`
3. Install jupyter lab and jupyter scheduler `pip install jupyter_scheduler jupyterlab`
4. Start jupyter lab `jupyter lab`
5. Create a notebook jobs and try to delete it by clicking on the delete button
<img width="1484" alt="Screen Shot 2023-03-31 at 10 27 27 AM" src="https://user-images.githubusercontent.com/46586569/229189272-daf15e47-2735-483e-a51d-6000e4c014c9.png">
6. Delete button not working as I cannot delete notebook jobs

## Expected behavior
Notebook jobs can be deleted by the Delete button

## Context

- Operating System and version: MacOS Version 12.6.3 (21G419)
- Browser and version: Safari Version 16.3 (17614.4.6.11.6, 17614)
- Jupyter Server version: 
```
jupyter --version 
Selected Jupyter core packages...
IPython          : 8.12.0
ipykernel        : 6.22.0
ipywidgets       : not installed
jupyter_client   : 8.1.0
jupyter_core     : 5.3.0
jupyter_server   : 2.5.0
jupyterlab       : 3.6.3
nbclient         : 0.7.2
nbconvert        : 7.2.10
nbformat         : 5.8.0
notebook         : 6.5.3
qtconsole        : not installed
traitlets        : 5.9.0
```

<details><summary>Troubleshoot Output</summary>
<pre>
❯ jupyter troubleshoot
$PATH:
        /Users/chaoneng/.nvm/versions/node/v14.21.3/bin
        /opt/homebrew/opt/ruby/bin
        /opt/homebrew/Caskroom/miniconda/base/envs/py310/bin
        /opt/homebrew/Caskroom/miniconda/base/condabin
        /opt/homebrew/bin
        /opt/homebrew/sbin
        /usr/local/bin
        /usr/bin
        /bin
        /usr/sbin
        /sbin
        /Users/chaoneng/.fig/bin
        /Users/chaoneng/.local/bin
        /Users/chaoneng/Library/Application Support/JetBrains/Toolbox/scripts
        /Users/chaoneng/.toolbox/bin
[...]

PR #383, opened 2023-06-12
Modal for delete job, stop job

JasonWeill:
Fix for #364; alternative to #382.

Fixes Safari compatibility by deleting the two-phase delete and stop buttons (icon becomes button, which in Safari is not properly focused) with icon buttons that spawn modal dialogs using MUI dialogs.

delete-with-confirm


andrii-i (changes requested):
While we wanted to avoid use of modal in the list view to make deleting large amount of jobs in the list view easier, making Delete button work on Safari is a priority. Thank you for taking care of this. 

While overall this works and looks good, please find a suggestion on better typing below.

JasonWeill:
Closing in favor of #385.

Issue #349, opened 2023-03-02
Unable to locate output files from a scheduled job run


AravindAmazon:
Hello team,

I have a job scheduled which will auto-run a basic Python script that writes a csv file output. The job runs successfully. However, am not able to locate the directory where the output file is scored.
The code line used for writing : temp.to_csv('trial_output.csv'), where temp is the data-frame variable.
When I use the same script in regular JupYter environment (outside JupYter lab), the csv file gets written successfully to the local JupYter environment folder. The issue appears to be happening only in the JupYter lab environment while using a scheduled job. Appreciate if someone can help (I use JupYter notebook via the AWS SageMaker interface).

Full-script:
import pandas as pd
temp = pd.read_csv("s3://<bucketname>/<objectname>")
temp.to_csv('trial_output.csv')

Overall purpose:
Require to auro-run case predictions on a daily basis (with a volume of atleast 10,000 predictions per day) and share a daily csv with business users (without any manual intervention)

Thanks,
Aravind

rubenvarela:
When I try recreating this, the file gets saved to the root folder of jupyter lab which maps to the location from where I ran the `jupyter-lab` command.

JasonWeill:
[ArchivingExecutionManager](https://github.com/jupyter-server/jupyter-scheduler/blob/5cd14b538a656e8fd2318dbbb687f2bf4da8fd37/jupyter_scheduler/executors.py#L176) archives the output files to a `.tar.gz` file, but it doesn't include files created as a side effect of running the notebook, as described in this issue.

This issue might be fixed by either modifying ArchivingExecutionManager or creating an alternate execution manager that gathers all output formats and all supporting files in and under the working directory, and saves them into an archive of some kind (`.zip` or `.tar.gz`).

JasonWeill:
Closing because #388 is merged.

JasonWeill:
To use the archiving scheduler, follow these instructions in the docs: https://jupyter-scheduler.readthedocs.io/en/latest/operators/index.html#example-capturing-side-effect-files

PR #388, opened 2023-06-28
Archiving all-files scheduler

JasonWeill:
WIP fix for #349.

Creates an `AllFilesArchivingExecutionManager` and `AllFilesArchivingScheduler` class, with the intention of archiving all output files and side effects on job runs.

You can use these classes by running:

```
jupyter lab \
  --SchedulerApp.scheduler_class=jupyter_scheduler.scheduler.AllFilesArchivingScheduler \
  --Scheduler.execution_manager_class=jupyter_scheduler.executors.AllFilesArchivingExecutionManager
```

After the user chooses to download output files from a job, the side effect files are unpacked to wherever the first output file is located.

In the screen shot below, after downloading the output of a job that ran `writefile.ipynb`, which produces the side effect file `pun.txt`, the side effect file is in the `jobs` subdirectory.

image


jupyter_scheduler/scheduler.py
```diff
@@ -709,8 +709,40 @@ def get_staging_paths(self, model: Union[DescribeJob, DescribeJobDefinition]) ->
...
+        id = model.job_id if isinstance(model, DescribeJob) else model.job_definition_id
+
+        for output_format in model.output_formats:
+            filename = create_output_filename(
+                model.input_filename, model.create_time, output_format
+            )
+            staging_paths[output_format] = filename
+
+        # Create an output zip file for automated runs
+        if isinstance(model, DescribeJob) and model.job_definition_id is not None:
```
3coins, line 741: I don't think we need to do any different here for run now jobs.

jupyter_scheduler/executors.py
```diff
@@ -221,3 +222,66 @@ def execute(self):
...
+        if job.parameters:
+            nb = add_parameters(nb, job.parameters)
+
+        ep = ExecutePreprocessor(
+            kernel_name=nb.metadata.kernelspec["name"],
+            store_widget_state=True,
+        )
+
+        # Get the directory of the input file
+        working_dir = os.path.dirname(os.path.abspath(self.staging_paths["input"]))
```
dlqqq, line 252: There's no need to call `os.path.abspath()` here, since the staging paths are all absolute paths.

jupyter_scheduler/executors.py
```diff
@@ -221,3 +222,66 @@ def execute(self):
...
+
+class AllFilesArchivingExecutionManager(DefaultExecutionManager):
+    """Execution manager that archives all output files in and under the
+    output directory into a single archive file
+
+    Notes
+    -----
+    Should be used along with :class:`~jupyter_scheduler.scheduler.AllFilesArchivingScheduler`
+    as the `scheduler_class` during jupyter server start.
+    """
```
dlqqq, line 235: There is a lot of similar code to `ArchivingExecutionManager`. I found that this made the code much more difficult to read, as I had to switch between the two implementations to discern the differences in behavior. I would highly encourage `AllFilesArchivingExecutionManager` to inherit from `ArchivingExecutionManager` and use both parent private methods and `super().execute()` to inherit logic and minimize code duplication.

jupyter_scheduler/job_files_manager.py
```diff
@@ -65,19 +65,44 @@ def generate_filepaths(self):
...
                     except Exception as e:
                         pass
+                if side_effect_file_directory is not None:
+                    # Extract all files in the tar.gz file,
+                    # other than the output formats listed above
+                    tar.extractall(
+                        side_effect_file_directory,
+                        members=self.exclude_output_filepaths(tar, filepaths),
+                        filter="data",
+                    )
```
dlqqq, line 105: The TGZ contains a flattened archive of the entire job staging directory at `<staging-dir>/<job-id>`, which also includes the input and output files. I think there's a significant opportunity to simplify the implementation here. Rather than extracting the input and output files from the tar, and then extracting the rest of the files separately via `tar.extractall(members=...)`, why not just call `tar.extractall()` with `members` unspecified?

This strategy would allow you to delete the `exclude_output_filepaths()` method and the `for` block above calling `tar.extractfile()` on each file type listed in `output_formats + ["input"]`.

jupyter_scheduler/executors.py
```diff
@@ -221,3 +222,66 @@ def execute(self):
...
+                for root, dirs, files in os.walk(working_dir):
+                    for file in files:
+                        with open(os.path.join(root, file)) as f:
+                            output = f.read()
+                        data = bytes(output, "utf-8")
+                        source_f = io.BytesIO(initial_bytes=data)
+                        # TODO: Include relative path?
+                        info = tarfile.TarInfo(file)
+                        info.size = len(data)
+                        tar.addfile(info, source_f)
```
dlqqq, line 283: 1. We should decide on whether we're OK with flattening all the side effect files and not preserving their directory structure. This will be a backwards-incompatible change, so we will want alignment on this.

2. The current implementation doesn't clean up side effect files generated under `<staging-dir>/<job-id>`. If we're OK with flattening all the side effect files, you could instead run the notebook under a temporary dir `<staging-dir>/<job-id>/.run` and delete all those files after building the tarball:

```py
working_dir = os.path.dirname(self.staging_paths["input"])
run_dir = os.path.join(working_dir, ".run")

        try:
            ep.preprocess(nb, {"metadata": {"path": run_dir }})
        except CellExecutionError as e:
            pass
        finally:
            # ... shutil.rmtree(run_dir)
```

I would also rename `working_dir` to `local_staging_dir` for clarity.

jupyter_scheduler/scheduler.py
```diff
@@ -709,8 +709,42 @@ def get_staging_paths(self, model: Union[DescribeJob, DescribeJobDefinition]) ->
...
-        staging_paths[output_format] = os.path.join(self.staging_path, model.job_id, filename)
-        staging_paths["input"] = os.path.join(self.staging_path, model.job_id, model.input_filename)
+        staging_paths[output_format] = os.path.join(self.staging_path, id, filename)
+        staging_paths["input"] = os.path.join(self.staging_path, id, model.input_filename)
+
+        return staging_paths
+
+
+class AllFilesArchivingScheduler(Scheduler):
+    """Scheduler that captures all files in output directory in an archive."""
```
dlqqq, line 719: This class appears to be logically identical to `ArchivingScheduler`, with two exceptions:

1. The default `execution_manager_class` is set to `AllFilesArchivingExecutionManager`

2. Minor readability changes right before the `return` statement.

I would instead apply the readability changes to `ArchivingScheduler`, have `AllFilesArchivingScheduler` inherit from that, and just override the trait's default by leaving the definition there. Then you can delete the duplicate `get_staging_paths()` implementation.

If I'm wrong about the two methods being logically identical, please see if you can use inheritance to minimize code duplication.

dlqqq (changes requested):
@JasonWeill Thank you for working on this! Left some feedback.

jupyter_scheduler/job_files_manager.py
```diff
@@ -65,19 +65,44 @@ def generate_filepaths(self):
             if not os.path.exists(output_filepath) or self.redownload:
                 yield input_filepath, output_filepath
 
+    def exclude_output_filepaths(self, members, filepaths):
+        output_filepaths = [filepath[1] for filepath in filepaths]
+        for tarinfo in members:
+            if tarinfo.name not in output_filepaths:
+                yield tarinfo
```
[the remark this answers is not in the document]
dlqqq: This may not be necessary; see https://github.com/jupyter-server/jupyter-scheduler/pull/388#discussion_r1289281656

dlqqq (changes requested):
There is also a failing unit test blocking CI:

```
FAILED jupyter_scheduler/tests/test_job_files_manager.py::test_downloader_download[output_formats1-output_filenames1-staging_paths1-/home/runner/work/jupyter-scheduler/jupyter-scheduler/jupyter_scheduler/tests/test_files_output-False] - AssertionError: assert False
 +  where False = <function exists at 0x7feb182f7420>('/home/runner/work/jupyter-scheduler/jupyter-scheduler/jupyter_scheduler/tests/test_files_output/helloworld-out.ipynb')
 +    where <function exists at 0x7feb182f7420> = <module 'posixpath' (frozen)>.exists
 +      where <module 'posixpath' (frozen)> = os.path
```

No need for an urgent fix here, since CI is failing on `main` anyways until we merge #417. However, this definitely should be fixed after rebasing and before merging.

jupyter_scheduler/job_files_manager.py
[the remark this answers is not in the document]
JasonWeill: My choice of the `members` name comes from Python's `tarfile` examples: https://docs.python.org/3/library/tarfile.html#examples

I'm also OK with simplifying the logic if the more specific function I wrote is not necessary.

JasonWeill:
Simplifies archiving and unarchiving logic per @dlqqq. Seeking external feedback about having these new classes replace the old `Archiving*` classes; if there's no opposition, I can work on this tomorrow. Also, we need to document these new classes and how to use them.

dlqqq (approved):
Awesome work! 🎉

Issue #390, opened 2023-07-03
Not compatible with Pydantic 2


JasonWeill:
[Pydantic 2.0](https://github.com/pydantic/pydantic/releases/tag/v2.0) was released on June 30, 2023. It causes errors when validating the results of server queries in the Python code:

* `@root_validator` is deprecated. We are advised to migrate to @model_validator, or to add `skip_on_failure=True` to the `@root_validator` call.
* `from_orm` is deprecated; we need to pass `from_attributes=True` in the model config to continue to use it in the short term, and remove reference to it eventually.

Issue #380, opened 2023-06-06
JupyterLab 4 compatibility


JasonWeill:
Make sure that Jupyter Scheduler works with JupyterLab 4.0.0 or later. Update dependencies and documentation.

JasonWeill:
Possibly relevant Discourse thread about targeting both versions: https://discourse.jupyter.org/t/jlab-extensions-that-target-both-jlab3-and-jlab4/20100

JasonWeill:
When I right-click on a file in the file browser, the "Create Job" option doesn't spawn the create form. Using the toolbar button in an open notebook does cause the form to appear.

PR #391, opened 2023-07-03
Pins Pydantic to version 1, adds Python 3.11

JasonWeill:
Mitigation for #390 (not compatible with Pydantic 2) and partial fix for #380 (JupyterLab 4 compatibility). Updates dependencies to ensure that only Pydantic v1 is used, for now. Adds Python 3.11 compatibility (tested and working).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,12 +26,13 @@ classifiers = [
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
     "jupyter_server>=1.6,<3",
     "traitlets",
     "nbconvert",
-    "pydantic",
+    "pydantic>=1,<2",
     "sqlalchemy<2",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -26,12 +26,13 @@ classifiers = [
...
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
     "jupyter_server>=1.6,<3",
     "traitlets",
     "nbconvert",
-    "pydantic",
+    "pydantic>=1,<2",
```
dlqqq, line 35: I would definitely leverage the compatible release operator `~=` as much as possible. The only quirk of this operator is that it behaves differently depending on how many version components you specify in the operand:

- `~=1.2` == `>=1.2.0,<2.0.0`
- `~=1.2.3` == `>=1.2.3,<1.3.0`

Here, we want `pydantic~=1.10`.

dlqqq (changes requested):

dlqqq (approved):

PR #402, opened 2023-07-19
Upgrade to JupyterLab 4

### LINKED ISSUE #380: JupyterLab 4 compatibility

dlqqq:
# Description

- Closes #380

This will have to be in a new major release (v2), since it is a breaking change for existing JupyterLab 3 users. However, we still intend to backport pertinent features and bug fixes back to v1.x at our discretion.

Surprisingly, the only necessary changes to get our code to build with JL4 were the following:

- Set `skipLibCheck: true` in `tsconfig.json`
- One-line change to `src/index.tsx`

These are implemented in https://github.com/jupyter-server/jupyter-scheduler/commit/51c551abd6c9544dcef635f5eca6ddf9d2f86c9a.

This begs the question of whether we should instead write the code in a manner that builds with both JupyterLab 3 & 4, and release two distributables: `jupyter_scheduler@2.0.0` (JL4) and `jupyter_scheduler_jl3@2.0.0` (JL3). Our team is currently discussing this at the moment.


JasonWeill:
Good work! The change to the iterator value (as seen in the file browser method you modified) was the only required change I saw locally, as well. I couldn't get both JL3 and JL4 to work in the same code without breaking all sorts of TypeScript type protections. I checked out and built your code locally and found that it worked as expected with Lab 4.0.3.

dlqqq:
please update playwright snapshots

dlqqq:
please update playwright snapshots

andrii-i (dismissed):
Good job! Please find some comments below. 

While best practices generally encourage making PRs as focused as possible and not mixing different types of changes, I don't see it as a big problem here, let's just be mindful of it and update the PR name to something like "Upgrade to JupyterLab 4, rework SchedulerHelper test class" for the sake of project history clarity.

andrii-i (approved):

PR #416, opened 2023-08-08
Server-side support for event driven notebook execution

andrii-i:


jupyter_scheduler/models.py
```diff
@@ -289,3 +306,9 @@ class JobFeature(str, Enum):
     output_filename_template = "output_filename_template"
     stop_job = "stop_job"
     delete_job = "delete_job"
+
+
+class Event(BaseModel):
+    event_id: str
+    event_type: str
+    parameters: Dict[str, str]
```
3coins, line 314: Should we have the type here as `Dict[str, Any]`?

jupyter_scheduler/orm.py
```diff
@@ -98,6 +98,7 @@ class Job(CommonColumns, Base):
     url = Column(String(256), default=generate_jobs_url)
     pid = Column(Integer)
     idempotency_token = Column(String(256))
+    triggered_by = Column(JsonType(1024), nullable=True)
```
3coins, line 101: See comments above about the naming.

jupyter_scheduler/orm.py
```diff
@@ -108,6 +109,7 @@ class JobDefinition(CommonColumns, Base):
     url = Column(String(256), default=generate_job_definitions_url)
     create_time = Column(Integer, default=get_utc_timestamp)
     active = Column(Boolean, default=True)
+    on_events = Column(JsonType(1024), default=list)
```
3coins, line 112: See comments above about naming.

3coins (commented):
@andrii-i 
Thanks for working on this. Left some comments about naming of events attribute.

jupyter_scheduler/models.py
andrii-i: Yes, good catch, thank you

andrii-i:
Thank you for the review @3coins

PR #417, opened 2023-08-09
Fix CI, run lint, reduce end-to-end tests flakiness

### LINKED ISSUE #399: run lint in CI
### Problem

Lint errors aren't being caught in CI: see https://github.com/jupyter-server/jupyter-scheduler/pull/387 immediately prior to my commits.

### Proposed Solution

Add a lint workflow to CI.

### LINKED ISSUE #413: Reduce E2E snapshot flakiness
## Description

The E2E test job is currently failing on `main`: https://github.com/jupyter-server/jupyter-scheduler/actions/runs/5670886374

Looking at the test report, there are two failures:

1. "shows notebook toolbar button" -- fails because the debugger is sometimes on, sometimes off. This test should use a more narrow locator that captures exclusively the left half of the notebook toolbar, rather than the entire main area widget.

Screenshot 2023-07-26 at 9 14 50 AM

2. "shows newly created job in list view" -- the content of the table is shifted to the right by a few pixels. I'm not sure what's happening here; it could be an upstream CSS change the affects the visual layout. I think just updating the snapshot and making sure the test isn't flaky is sufficient.

Screenshot 2023-07-26 at 9 16 58 AM

andrii-i:
- Fixes CI by pinning jupyterlab version for end-to-end tests and snapshot update jobs
- Fixes #413 by always setting `create_time` to `1` (`Jan 1, 1970 12:00 AM`)
- Fixes #399. Lint problems will now result in a `build` CI workflow failure. See [this](https://github.com/jupyter-server/jupyter-scheduler/actions/runs/5828293778/job/15805772661?pr=417) run for an example of a failed CI run
Screenshot 2023-08-10 at 7 34 11 PM


ui-tests/tests/jupyter_scheduler.spec.ts
```diff
@@ -66,9 +73,9 @@ test.describe('Jupyter Scheduler', () => {
     await scheduler.createNotebook();
     await scheduler.createJobFromFilebrowser();
 
-    const timeStamp = scheduler.timestampLocator;
     await scheduler.assertSnapshot(FILENAMES.LIST_VIEW, {
-      mask: [timeStamp]
+      mask: [scheduler.timestamp],
+      maxDiffPixels: 420
```
dlqqq, line 78: I would prefer us never using `maxDiffPixels`, as that kind of defeats the point of snapshot testing IMO. I have two questions:
1. What is causing the flakiness? What elements specifically are causing the layout to change test-to-test?

2. Is it possible to use a more narrow locator or a wider set of masks that doesn't cause any test flakiness?

dlqqq (changes requested):

ui-tests/tests/jupyter_scheduler.spec.ts
andrii-i: I agree that having `maxDiffPixels` is not sustainable and should be fixed in this or following PRs. 

Advanced Table causes it. All columns except job name shift left-right a couple of pixels between runs. Below is a heatmap of the difference. I have not caught this programmatically but I can reproduce a similar shift by editing content of the table cells and making it overflow. 


We could mask all columns in the table (headers and values) except of the job name or hardcode the values but IMO it also kinda defeats the purpose of the test.

list-view-diff

dlqqq: See if you can set the `create_time` timestamp retrieved from ListJobs and ListJobDefinitions to a fixed value via PlayWright: https://playwright.dev/docs/network#modify-responses

dlqqq: I'd recommend a default of `0` (Jan 1, 1970 12:00 AM).

andrii-i: Ok`page.route` was finicky but I made it work. Interestingly `0` parses to empty value and `1` parses to `Jan 1, 1970 12:00 AM)`
Screenshot 2023-08-10 at 6 43 44 PM

dlqqq (approved):
Awesome work! Very happy we are able to remove the mask on the ListJobs view.

PR #419, opened 2023-08-11
Avoids "filter" option in tarfile

JasonWeill:
Removes the `filter` option used in `tarfile.extractall()`, which is only compatible with Python 3.11.4 and newer.

Follow-up to #388, included in #418 (this does _not_ need backporting to 1.x)


dlqqq (approved):
CI failures should be fixed by #424.

PR #424, opened 2023-08-15
Fix JFM tests

### LINKED ISSUE #420: Pytest failure in main (after #388 was merged) and #418
## Description

- Build CI workflow fails due to Pytest `jupyter_scheduler/tests/test_job_files_manager.py:125` failure after changes in main (after #388 was merged) and #418. 

## Reproduce

See CI run : https://github.com/jupyter-server/jupyter-scheduler/actions/runs/5837424752/job/15832930590

dlqqq:
- Closes #420.

It looks like the failure was the result of two things:

1. `shutil.rmtree()` isn't guaranteed to be synchronous, so it's necessary to implement a short poll until the directory is deleted from the filesystem.

2. We weren't automatically creating the output directory in the case of TGZ.

### DIFF
--- a/jupyter_scheduler/tests/test_job_files_manager.py
+++ b/jupyter_scheduler/tests/test_job_files_manager.py
@@ -67,6 +67,7 @@ OUTPUTS_DIR = os.path.join(HERE, "test_files_output")
 def clear_outputs_dir():
     yield
     shutil.rmtree(OUTPUTS_DIR)
+    # rmtree() is not synchronous; wait until it has finished running
     while os.path.isdir(OUTPUTS_DIR):
         time.sleep(0.01)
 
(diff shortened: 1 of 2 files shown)


jupyter_scheduler/tests/test_job_files_manager.py
```diff
@@ -68,6 +67,8 @@ async def test_copy_from_staging():
 def clear_outputs_dir():
     yield
     shutil.rmtree(OUTPUTS_DIR)
+    while os.path.isdir(OUTPUTS_DIR):
```
JasonWeill, line 70:
```suggestion
    # rmtree() is not synchronous; wait until it has finished running
    while os.path.isdir(OUTPUTS_DIR):
```

JasonWeill (approved):
Thanks again for your help! One minor comment, otherwise good to go

PR #438, opened 2023-10-03
Fix RTD and E2E CI workflows

andrii-i:
- Fixes ReadTheDocs build process (error example: https://readthedocs.org/projects/jupyter-scheduler/builds/22111095/) by adding readthedocs configuration file v2

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -25,6 +25,8 @@ jobs:
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
 
+      # when updating this version, make sure to update ui-tests/package.json
+      # and the lockfile accordingly.
       - name: Install extension dependencies
         run: python -m pip install -U jupyterlab==4.0.6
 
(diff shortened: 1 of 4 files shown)


dlqqq:
I just rebased and squashed away the commit from the pre-commit bot. Will see if E2E tests pass after rerunning the pipeline.

dlqqq:
Ah, this looks like an upstream change in Playwright that somehow slipped by, even when fixing `jupyterlab==4.0.3` in our tests. See here: https://github.com/jupyterlab/jupyterlab/issues/15116

JasonWeill:
Not just our repo is affected; I had to upgrade readthedocs config in https://github.com/jupyter/jupyter/pull/698/ as well

dlqqq:
Resolved this issue. Here was my thought process (this may be helpful to @andrii-i):
 
The PR that backports a fix for this issue to 4.0.x is https://github.com/jupyterlab/jupyterlab/pull/15118. Looking at the PR, we see it was merged in as commit `f73ea9acc67a5dd581e64b400018d3373d3676dc`. Then, I ran:

```
git tag --contains f73ea9acc67a5dd581e64b400018d3373d3676dc
```

to see that it was first included in the v4.0.6 release. This indicated to me that upgrading to v4.0.6 in our tests would be solve this CI failure. After making sure to also set this patch version for Galata and updating `ui-tests/yarn.lock`, I was able to get CI passing again.

dlqqq (approved):

andrii-i:
Thank you @dlqqq

PR #444, opened 2023-10-12
[1.x] Fix RTD CI workflow

andrii-i:
Backport #438 to 1.x (but skip updating jupyterlab version to 4.0.6 as it is not relevant for 1.x)


andrii-i:
Thank you @dlqqq

PR #446, opened 2023-10-20
Add Notifications support

andrii-i:


andrii-i:
@Zsailer would love to get your feedback on this PR adding notifications support to Jupyter Scheduler. Please add anyone else who might be relevant.

src/components/notification-detail.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { LabeledValue } from '../components/labeled-value';
+
+interface INotificationItemProps {
+  label: string;
+  value: string | boolean;
+}
+
+interface INotificationDetailsProps {
+  notification: Scheduler.INotification;
+}
```
dlqqq, line 15: nit: [use `type` instead of `interface`](https://www.totaltypescript.com/type-vs-interface-which-should-you-use#default-to-type-not-interface). Interfaces are just more characters to type and have some surprising side-effects like declaration merging.

This is personal preference; change is encouraged but not necessary.

src/components/notification-detail.tsx
```diff
@@ -0,0 +1,69 @@
...
+            />
+          ))}
+          <FormLabel component="legend">{trans.__('Events')}</FormLabel>
+          {notification.events.map((event, idx) => (
+            <NotificationItem
+              key={idx}
+              label={trans.__(`Event ${idx + 1}`)}
+              value={event}
+            />
+          ))}
```
dlqqq, line 60: The UI here needs work. For the emails, you should probably use a table, list of badges, or something else that doesn't take up a block element for every email. This gets hard to read after maybe 5 emails.

For the events, recall that the number of events is bounded to 3 presently. So you can just comma-join all the events and use a single field to indicate that.

src/components/notification-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+    <Stack size={2}>
+      <InputLabel>{trans.__('Notification')}</InputLabel>
+      <FormControlLabel
+        control={
+          <Switch
+            checked={enableNotification}
+            onChange={enableNotificationChange}
+          />
+        }
+        label={trans.__('Enable Notification')}
```
dlqqq, line 139: Don't use Title Case, but instead Sentence case: `'Enable notification'`. This matches the convention for other field labels.

dlqqq: Do this for the other fields as well, like "Send To" and "Include Output".

src/components/notification-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+    </Stack>
+  );
+}
+
+const NotificationEventsSelect: React.FC<{
+  id: string;
+  availableEvents: string[];
+  selectChange: (e: SelectChangeEvent<string>) => void;
+  disabled: boolean;
+}> = ({ id, availableEvents, selectChange, disabled }) => {
```
dlqqq, line 183: 1. Why define these components via arrow functions instead of plain old functions?  Compare the two in readability:

```ts
const Component: React.FC<ComponentProps> = (props) => { /* ... */ }

// v.s.

function Component(props: ComponentProps) { /* ... */ }
```

It's also worth noting that functions are [hoisted](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) via const arrow functions are not. This is a pretty useful property in general, and is another reason why I prefer plain old functions.

2. Please extract the props type declaration into a separate type rather than doing it inline. Complex inline types are difficult to parse, and the lack of a name makes their purpose obscure until studied carefully.
3. nit: I see that you're using ES6 destructuring to avoid having to prefix `props.*` to every prop. While I totally understand the motivation to type less, that prefix is a helpful indicator that "this expression is a React prop" and distinguishes it from local variables which are generally less important helpers.

dlqqq: Same goes for `SelectedEventsChips`.

src/components/notification-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+        label={trans.__('Include Output')}
+      />
+    </Stack>
+  );
+}
+
+const NotificationEventsSelect: React.FC<{
+  id: string;
+  availableEvents: string[];
+  selectChange: (e: SelectChangeEvent<string>) => void;
```
dlqqq, line 181: Also, for input components, I'd prefer us to stick to the convention of always labeling the input state & change handler as `value` and `onChange` respectively.

src/components/notification-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+import { ICreateJobModel } from '../model';
+import Select, { SelectChangeEvent } from '@mui/material/Select';
+import { Stack } from './stack';
+import { useTranslator } from '../hooks';
+
+interface INotificationPickerProps {
+  notificationEvents: string[];
+  id: string;
+  model: ICreateJobModel;
+  handleModelChange: (model: ICreateJobModel) => void;
```
dlqqq, line 23: See if you can make the interface smaller by only passing `model.notification`, and similarly have the change handler only accept changes to that sub-object. Doing something like

```
        modelChange({
          ...model,
          notification: {
            ...model.notification,
            selectedEvents: updatedEvents
          }
        });
```

on every change is risky, since if a developer later forgets the `...model` line, then the form breaks. Keeping the props interface for a React component as small as possible is a great practice to follow.

src/components/notification-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+  );
+  const [enableNotification, setEnableNotification] = useState<boolean>(
+    model.notification?.enableNotification ?? true
+  );
+  const [sendToInput, setSendToInput] = useState<string>(
+    model.notification?.sendTo?.join(', ') || ''
+  );
+  const [includeOutput, setIncludeOutput] = useState<boolean>(
+    model.notification?.includeOutput || false
+  );
```
dlqqq, line 44: This appears to all be duplicate state. You are defining local state variables that don't really do anything except sync with the corresponding property on `model.notification`. This is very dangerous, because then you can have a mismatch between what the UI shows (from the local state) and what is actually being sent to the backend (from the model state).

Unless there is a good reason to have the local and model state be separate, this component should access the state via `props.model.notification` and change the state via `props.handleModelChange`. This will be much easier if you narrow the scope of the props to only concern the notifications, as I recommended previously.

src/mainviews/create-job.tsx
```diff
@@ -121,12 +122,18 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
           ...props.model,
           environment: envList[0].name,
           computeType: newComputeType,
           outputFormats: outputFormats
-        });
+        };
+
+        if (envList[0].notifications_enabled && !props.model.notification) {
+          newModel.notification = { enableNotification: true };
+        }
```
dlqqq, line 134: Similar to my previous comment about duplicate state, I believe this extra attribute can also be removed since we should be able to derive this from the other fields on the model when we need it. That way you won't have to worry about duplicate state getting out-of-sync.

dlqqq (changes requested):

jupyter_scheduler/models.py
```diff
@@ -26,6 +56,8 @@ class RuntimeEnvironment(BaseModel):
     compute_types: Optional[List[str]]
     default_compute_type: Optional[str]  # Should be a member of the compute_types list
     utc_only: Optional[bool]
+    notifications_enabled: bool = False
+    notification_events: List[Type[NotificationEvent]] = [e for e in NotificationEvent]
```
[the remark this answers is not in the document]
andrii-i: Good point taking in account that `notifications_enabled` defaults to `False`

jupyter_scheduler/models.py
```diff
@@ -11,6 +11,36 @@
...
+        FAILURE (str): Sent on job failure.
+        STOPPED (str): Sent when a job is manually stopped.
+    """
+
+    SUCCESS = "Success"
+    FAILURE = "Failure"
+    STOPPED = "Stopped"
+
+
+class Notification(BaseModel):
```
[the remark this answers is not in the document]
andrii-i: I thought about this + potential confusion with [Notification class in JupyterLab](https://github.com/jupyterlab/jupyterlab/blob/main/jupyterlab/handlers/announcements.py) but still decided to opt for a short name that is clear enough given it's use within Scheduler extension (+ I thought alternatives I came up with were not adding much clarity).

I'm ok to use a more verbose name for the sake of clarity but let's consider options. Problem with `ScheduledNotifications` is that only notebook execution is scheduled. We don't schedule notifications, we define to whom and based on which event/status change they should be sent. `PendingNotifications` to me means a batch of notification objects that are waiting to be sent.

Taking this all into account, `NotificationsSettings` seems like a good name because it does a good job explaining what this class is (necessary data / settings / configuration to send multiple notifications later). What do you think?

jupyter_scheduler/orm.py
```diff
@@ -108,6 +109,7 @@ class JobDefinition(CommonColumns, Base):
     url = Column(String(256), default=generate_job_definitions_url)
     create_time = Column(Integer, default=get_utc_timestamp)
     active = Column(Boolean, default=True)
+    notification = Column(JsonType(1024), nullable=True)
```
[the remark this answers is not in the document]
andrii-i: Thanks for pointing this out. It would have been best to address this during the design review, but as discussed there was not very much time to dive deep into all the aspects of the design doc.

src/components/notification-picker.tsx
andrii-i: 1. Using arrow functions allows to destructure props within the function declaration making code more concise and readable (vs destructuring in a separate line; if we would like to avoid destructuring props in our codebase as you suggest, this is less of an advantage). Using `React.FC<TYPE>` also provides explicit and standardized function signature that does not rely on a convention and manual typing of `props` which IMO makes code easier to read and write. 

I don't have a hard preference here and think that both approaches have merit as well as pros and cons. It seems like you do so I'm ok to use functions instead. Interestingly, Netlify's React Typescript Cheatsheet agrees with you and expresses opinion that React.FC might be unnecessary: https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/

2. I think this is a question of a right balance between verbosity and clarity. In your opinion, should props type declarations always be separate types/interfaces? 

3. I think this is also a question of balance between verbosity and clarity.

Don't have a hard preference on 2 or 3 as well so will update in line with your comments.

src/components/notification-detail.tsx
andrii-i: Thanks for the link. I'm actually defaulting to interfaces partly based on typescript handbook [recommendation](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) that article you link references and disagrees with.

src/components/notification-picker.tsx
andrii-i: Should we use ESLint `func-style` rule allowing only declaration-type functions and explicitly disallowing arrow functions to formalize this informal preference for plain old functions into a formal requirement? This way we won't have to discuss this in the future. I can do this when time allows.

https://eslint.org/docs/latest/rules/func-style#options

Zsailer:
@akshaychitneni

akshaychitneni:
@andrii-i thanks for adding the notification support. I think it would also be helpful to support multiple notification types like email, slack etc and encapsulate in the schema. wdyt?

jupyter_scheduler/models.py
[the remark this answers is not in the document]
dlqqq: `NotificationConfig` would be good. 👍

src/components/notification-picker.tsx
dlqqq: Arrow functions allow you to define an inline function very concisely, so they still have their uses in change handlers for example. So I don't think it's necessary to add this ESLint rule.

andrii-i:
@akshaychitneni, thank you for your feedback. When creating this PR, I was envisioning that third parties implementing the notification delivery mechanisms would independently decide on the notification types. However, I see the potential benefit of directly supporting various notification types backed by a schema.

I’d like to better understand your vision: imagine a scenario where users can pick anywhere from zero to several notification types for each job. Would the JSON schema fields then be meant more as informational aids to help guide their choices? Or, are you envisioning a setup where users would be actively filling out specific fields for each notification type, as laid out in the JSON schema?

Zsailer:
> When creating this PR, I was envisioning that third parties implementing the notification delivery mechanisms would independently decide on the notification types

I haven't had a chance to dive into the details of this code, but one of the general challenges we've faced with the scheduler plugin is that it's difficult to extend without re-writing a significant portion of the server extension and client plugin. 

How would one go about defining their own notification type in this implementation?  

> Or, are you envisioning a setup where users would be actively filling out specific fields for each notification type, as laid out in the JSON schema?

Ideally, I'd love to have the job definition specified by a JSON schema defined _before_ the server starts. The job definitions coming in from the client can be automatically validated by the backend (via pydantic, or whatever tool you like) using this schema and the forms/views in the frontend are automatically generated using the schema.

That way, this plugin can be extended/customized by the "user"—most likely this means the person deploying JupyterLab to a bunch of users who want to submit Notebook jobs—simply by extending the JSON schema.

Zsailer:
I'm thinking of something like https://jsonforms.io/

If the job definition can be specified by a (JSON) schema, this extension becomes _extendable_.

andrii-i:
> I haven't had a chance to dive into the details of this code, but one of the general challenges we've faced with the scheduler plugin is that it's difficult to extend without re-writing a significant portion of the server extension and client plugin.
>
>How would one go about defining their own notification type in this implementation?

As it stands now, extension does not provide mechanisms for delivery of notifications or for defining additional properties such as notification types. `RuntimeEnvironment` data model has hardcoded notifications-related fields meant for configuration, `notifications_enabled` and `notification_events`. We could change `RuntimeEnvironment` data model and add new field, for example `notification_type` or a field with schema. 

To set `RuntimeEnvironment` fields (for example, set `notifications_enabled` or provide an array of `notification_events`), 3rd parties are supposed to:
1. Subclass `CondaEnvironmentManager` and override `list_environments`  method so that it would create/return the modified `RuntimeEnvironment`  (https://github.com/jupyter-server/jupyter-scheduler/blob/main/jupyter_scheduler/environments.py#L29)
2. Specify custom environment manager as the value for `environment_manager_class` trait (https://github.com/jupyter-server/jupyter-scheduler/blob/main/jupyter_scheduler/extension.py#L48C7-L48C7)

In earlier iterations of the design I also had json schema field in the `RuntimeEnvironment` that allowed to define generic fields for notifications UI but we decided against proceeding with it at the time.

andrii-i:
> I haven't had a chance to dive into the details of this code, but one of the general challenges we've faced with the scheduler plugin is that it's difficult to extend without re-writing a significant portion of the server extension and client plugin.
>
>How would one go about defining their own notification type in this implementation?

@Zsailer Thank you for looking into this. As it stands now, extension does not provide mechanisms for delivery of notifications or for defining additional properties such as notification types. `RuntimeEnvironment` data model has hardcoded notifications-related fields meant for configuration, `notifications_enabled` and `notification_events`. We could change `RuntimeEnvironment` data model and add new field, for example `notification_type` or a field with schema. 

To set `RuntimeEnvironment` fields (for example, set `notifications_enabled` or provide an array of `notification_events`), 3rd parties are supposed to:
1. Subclass `CondaEnvironmentManager` and override `list_environments`  method so that it would create/return the modified `RuntimeEnvironment`  (https://github.com/jupyter-server/jupyter-scheduler/blob/main/jupyter_scheduler/environments.py#L29)
2. Specify custom environment manager as the value for `environment_manager_class` trait (https://github.com/jupyter-server/jupyter-scheduler/blob/main/jupyter_scheduler/extension.py#L48C7-L48C7)

In earlier iterations of the design I also had json schema field in the `RuntimeEnvironment` that allowed to define generic fields for notifications UI but we decided against proceeding with it at the time.

jupyter_scheduler/orm.py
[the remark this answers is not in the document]
andrii-i: Added a separate table for `NotificationsConfig` (note that eager loading relevant `NotificationsConfig` object is required to prevent `DetachedInstanceError` that occurs when Job instance detached from the session tries to access notifications config). There is no native SQLAlchemy data type that covers lists of strings with necessary granularity (both number of items in the list and size of each item) so I opted for enforcing constraints on application level through [pydantic validators](https://github.com/jupyter-server/jupyter-scheduler/pull/446/files#diff-20dd20fee14818d8deb4f44eb356356178e132f8b838a28c0a4027a0e73ac073R43-R67).

src/components/notification-detail.tsx
andrii-i: Considered different options. Decided to go with comma-separated string of values to display both send to symbols and notification events as it looks most consistent with how other values are displayed in details view:
Screenshot 2023-11-02 at 9 27 53 AM

In my opinion, if we for example use chips to display send to symbols (or send to and events), send to field becomes visually and logically more important than other fields. This is not always the case so always visually emphasizing send to field decreases readability:
Screenshot 2023-11-02 at 9 25 29 AM
Screenshot 2023-11-02 at 9 21 45 AM


Overall, I agree that details view design could use further improvement but it needs to be redesigned as a whole to maintain visual consistency.

cflann:
I would love to see this feature added. Thank you @andrii-i

krassowski:
I agree it would be nice to have this feature. Of note [argo-jupyter-scheduler](https://github.com/nebari-dev/argo-jupyter-scheduler) implements Slack notifications. If it could hook into the new API to offer consistent UI/UX that would be amazing.

PR #448, opened 2023-10-26
Added telemetry support.

3coins:
## Description
This PR adds support for telemetry and emits data for different events in the Scheduler components. A new plugin is provided that receives the event data. Implementors can disable the default plugin and provide their own implementation that will pass the telemetry data to their handler.

#### Advanced options expanded
```json
{
    "body": {
        "name": "org.jupyter.jupyter-scheduler.create-job.advanced-options.expand"
    },
    "timestamp": "2023-10-26T17:17:44.350Z"
}
```

#### Create job cancel button clicked
```json
{
    "body": {
        "name": "org.jupyter.jupyter-scheduler.create-job.cancel"
    },
    "timestamp": "2023-10-26T17:18:53.323Z"
}
```

#### Create job run on schedule option selected
```json
{
    "body": {
        "name": "org.jupyter.jupyter-scheduler.create-job.job-type.run-on-schedule"
    },
    "timestamp": "2023-10-26T17:19:28.917Z"
}
```

### Providing your implementation to process telemetry events
 Create a new handler and a plugin that provides the handler.
```python

const myTelemetryHandler: async (eventLog: Scheduler.IEventLog): Promise<void> => {
    // process the eventLog 
}

const myTelemetryPlugin: JupyterFrontEndPlugin<Scheduler.TelemetryHandler> = {
  id: '@jupyterlab/scheduler:MyTelemetryHandler',
  autoStart: true,
  provides: Scheduler.TelemetryHandler,
  activate: (app: JupyterFrontEnd) => {
    return myTelemetryHandler;
  }
};
```

Disable the default plugin in the `package.json` file of your extension package.

```json
"jupyterlab": {
    "extension": true,
    "disabledExtensions": [
      "@jupyterlab/scheduler:TelemetryHandler"
    ]
  }
```

### Attaching new events to telemetry handler
A new hook `useLogger` is available that could be accessed in any component.

```ts
const log = useLogger();
```

Any event handler can then call the `log` to pass in the event name.

```ts
log("job-detail.<my-event>")
```

### DIFF
--- a/src/mainviews/create-job.tsx
+++ b/src/mainviews/create-job.tsx
@@ -530,12 +530,18 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
             handleErrorsChange={setErrors}
           />
           <Accordion
-            telemetry-event-name={`create-job.advanced-options.${advancedOptionsExpanded ? 'expanded': 'collapsed'}`}
+            telemetry-event-name={`create-job.advanced-options.${
+              advancedOptionsExpanded ? 'expanded' : 'collapsed'
+            }`}
             defaultExpanded={false}
             expanded={advancedOptionsExpanded}
             onChange={(e: React.SyntheticEvent, expanded: boolean) => {
-              telemetryHandler(`create-job.advanced-options.${expanded ? 'expand': 'collapse'}`);
-              setAdvancedOptionsExpanded(expanded)
+              telemetryHandler(
+                `create-job.advanced-options.${
+                  expanded ? 'expand' : 'collapse'
+                }`
+              );
+              setAdvancedOptionsExpanded(expanded);
             }}
           >
             <AccordionSummary
@@ -580,8 +586,8 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
                 <Button
                   variant="outlined"
                   onClick={e => {
-                    telemetryHandler("create-job.cancel")
-                    props.showListView(JobsView.ListJobs)
+                    telemetryHandler('create-job.cancel');
+                    props.showListView(JobsView.ListJobs);
                   }}
                 >
                   {trans.__('Cancel')}
@@ -590,7 +596,7 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
                 <Button
                   variant="contained"
                   onClick={(e: React.MouseEvent) => {
-                    telemetryHandler("create-job.create")
+                    telemetryHandler('create-job.create');
                     submitForm(e);
                     return false;
                   }}
--- a/src/mainviews/create-job.tsx
+++ b/src/mainviews/create-job.tsx
@@ -3,8 +3,7 @@ import React, {
   useEffect,
   useMemo,
   useState,
-  useRef,
-  useContext
+  useRef
 } from 'react';
 
 import { Heading } from '../components/heading';
@@ -18,7 +17,7 @@ import {
 } from '../components/output-format-picker';
 import { ParametersPicker } from '../components/parameters-picker';
 import { Scheduler, SchedulerService } from '../handler';
-import { useTranslator } from '../hooks';
+import { useLogger, useTranslator } from '../hooks';
 import { ICreateJobModel, IJobParameter, JobsView } from '../model';
 import { Scheduler as SchedulerTokens } from '../tokens';
 import { NameError } from '../util/job-name-validation';
@@ -42,7 +41,6 @@ import {
 } from '@mui/material';
 
 import { Box, Stack } from '@mui/system';
-import { TelemetryContext } from '../context';
 
 export interface ICreateJobProps {
   model: ICreateJobModel;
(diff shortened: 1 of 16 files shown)


dlqqq (changes requested):
This is a great start! Awesome work. I left 2 pieces of minor feedback below, but also want to suggest a broader name change here. I want to first call your attention to how consumers are using the hook:

```
const telemetryHandler = useContext(TelemetryContext)
```

The hook return `telemetryHandler` is rather opaque since it doesn't tell us what this function does. Like yes, it handles telemetry, but that's not very precise. This name is also overused since there are identical/similar names elsewhere:
1. One with `telemetryHandler` as defined in `index.tsx`, and
2. One with `handleTelemetry` as defined in `src/notebook-jobs-panel.tsx`.

Telemetry isn't a verb. It is a noun which refers to the collection of data obtained through metrics, logs, and traces collectively, each of which has its own precise definition. What we're really doing when calling the hook return is _logging_.

I suggest the following changes in `context.ts`:
- `ITelemetryProcessor` => `Logger`
- `TelemetryContext` => `LogContext`
- `useContext(TelemetryContext)` => `useLogger()` (define a custom hook for this)

These naming changes allow for a much more concise and readable syntax for consumers:

```
const log = useLogger()

// instead of:

const telemetryHandler = useContext(TelemetryContext)
```

Here, it's immediately obvious that the hook return is a function which logs something. Then it makes sense that logs are handled by a `telemetryHandler` in `index.tsx` because logs are a part of telemetry.

The plugin and token names don't need to be changed significantly because their name is accurate; they are handling telemetry from the emitted logs. I would drop the `I` prefix since they are not interfaces, e.g. `ITelemetryHandler => TelemetryHandler`.

Finally, in `src/notebook-jobs-panel.tsx`, I would rename the `handleTelemetry()` method to just `log()`, since this method is the value of the `LoggerContext` and should be named accordingly.

src/notebook-jobs-panel.tsx
```diff
@@ -219,7 +233,7 @@ export class NotebookJobsPanel extends VDomRenderer<JobsModel> {
 
     return (
       <ThemeProvider theme={getJupyterLabTheme()}>
-        <TranslatorContext.Provider value={this._translator}>
+        <TelemetryContext.Provider value={this.handleTelemetry.bind(this)}>
```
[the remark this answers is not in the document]
3coins: Good catch! Will update this.

src/notebook-jobs-panel.tsx
```diff
@@ -122,6 +124,18 @@ export class NotebookJobsPanel extends VDomRenderer<JobsModel> {
     }
   };
 
+  handleTelemetry(eventName: string): void {
+    if (eventName) {
+      const eventLog = {
+        body: {
+          name: `org.jupyter.jupyter-scheduler.${eventName}`
+        },
+        timestamp: new Date()
```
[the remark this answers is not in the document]
3coins: Keeping it a date object keeps it flexible for the consumer to format it the way they like.

src/mainviews/create-job.tsx
```diff
@@ -528,9 +530,14 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+            onChange={(e: React.SyntheticEvent, expanded: boolean) => {
+              log(
+                `create-job.advanced-options.${
+                  expanded ? 'expand' : 'collapse'
+                }`
+              );
+              setAdvancedOptionsExpanded(expanded);
+            }}
           >
             <AccordionSummary
```
JasonWeill, line 542: Should we capture events when the accordion is expanded or collapsed?

JasonWeill (commented):
General comment; I was confused by the use of `log` and `GetLogger` for telemetry events. We use very similar names for functions that log things to stdout/stderr, for example. Could we use something like `logTelemetryEvent` or `telemetryEvent` to be more noticeably different?

src/mainviews/create-job.tsx
3coins: The log statement above this does exactly that. Did you mean to capture something else?

3coins:
@JasonWeill 

> I was confused by the use of `log` and `GetLogger` for telemetry events. 

See the comments from David's review explaining the rationale behind this.

src/mainviews/create-job.tsx
JasonWeill: My mistake; I was looking in the wrong place here.

JasonWeill:
While I agree that "telemetry" isn't a verb, the verb shouldn't just be "log" for telemetry events; that's ambiguous. I'd prefer that `log` be kept for stdout/stderr logging, and `logTelemetryEvent` or `logEvent` be used for telemetry event logging.

3coins:
@JasonWeill @dlqqq 
Made final updates for the names.

JasonWeill (approved):

dlqqq (approved):
Awesome work!

PR #450, opened 2023-10-30
Fix "event loop is already running" bug on Linux

dlqqq:
## Description

- Fixes #435
- See also #330

This PR fixes the dreaded "event loop is already running" bug. I've determined the root cause of this bug and will explain why these changes fix it.

## Context

First, I would like to describe the context:

- The executor is started as a separate process via the class `multiprocessing.Process`.
- The executor calls synchronous methods on the class `nbconvert.preprocessors:ExecutePreprocessor`, which are actually implemented by wrapping the async methods in a function named `run_sync()`, provided by `jupyter_core.utils`
  - `run_sync()`, as you might expect, takes a coroutine and returns a synchronous function that "does the same thing". For more specifics, it's best to refer to the code directly.

## First cause

When `nbconvert` calls `run_sync()`, it _sometimes_ throws an exception with the message "The event loop is already running".

## Root cause

- On all non-POSIX systems (e.g. Windows & MacOS), `multiprocessing` starts a process by spawning a new one from scratch.
- [However, only on POSIX systems (e.g. Linux), `multiprocessing` starts new processes by forking the current process.](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)
- [`asyncio.get_event_loop()` fails in forked child processes](https://github.com/python/cpython/issues/66285), as it returns the event loop of the parent process instead of the child process in which it was called.
  - This may be fixed in Python 3.12, but I do not have the time to verify. See: https://github.com/python/cpython/pull/99769

This means that on Linux, `asyncio.get_event_loop()` in a new process actually returns the event loop of the server process, which of course is always running (since it's running the server). This is why the exception is thrown. 😁 

## The fix

To fix this, we need to indicate to `multiprocessing` (MP) to only start processes via `spawn` instead of `fork`. Thankfully this can be done via MP contexts, as shown in this PR.

I also see that we are running the Downloader in a new process via MP. However, I did not implement the same changes there for 3 reasons:

1. The default Downloader (provided by the default JobFilesManager) works on Linux without any further changes.
2. Since the Downloader's interface only has synchronous methods, an implementer would have to go out of their way to call `asyncio.get_event_loop()` in the implementation for the same bug to appear.
3. Using `spawn` instead of `fork` does have a significant performance drawback, so unless there is reason to otherwise, we should stick with MP's default behavior.

## Next steps

I filed a new issue to add unit test coverage: https://github.com/jupyter-server/jupyter-scheduler/issues/453

PR #452, opened 2023-10-30
Migrate from hub to gh in workflows

dlqqq:
See https://github.com/actions/runner-images/issues/8531

I need this PR to update the snapshots in #450.


dlqqq:
This PR does not need to be backported to `1.x`, since the workflow as defined in `main` is used by default.

3coins (approved):

PR #454, opened 2023-10-30
[1.x] Fix "event loop is already running" bug on Linux

dlqqq:
Backports #450 to 1.x.


dlqqq:
please update playwright snapshots

dlqqq:
The above command fails because it's always using the workflow from `main`, which uses JupyterLab 4. I believe the only way to update snapshots in 1.x is to invoke the workflow manually in the "Actions" tab, which allows you to select the branch.

PR #457, opened 2023-10-31
Telemetry patch

3coins:
Moved the event logger handler up in the stack to capture launcher, context menu and notebook header events.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -72,7 +72,7 @@ const advancedOptions: JupyterFrontEndPlugin<Scheduler.IAdvancedOptions> = {
 const telemetryHandler = async (
   eventLog: Scheduler.IEventLog
 ): Promise<void> => {
-  console.log(JSON.stringify(eventLog, undefined, 4));
+  // console.log(JSON.stringify(eventLog, undefined, 4));
 };
 
 const telemetry: JupyterFrontEndPlugin<Scheduler.TelemetryHandler> = {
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -72,7 +72,7 @@ const advancedOptions: JupyterFrontEndPlugin<Scheduler.IAdvancedOptions> = {
 const telemetryHandler = async (
   eventLog: Scheduler.IEventLog
 ): Promise<void> => {
-  console.log(JSON.stringify(eventLog, undefined, 4));
+  // console.log(JSON.stringify(eventLog, undefined, 4));
 };
 
 const telemetry: JupyterFrontEndPlugin<Scheduler.TelemetryHandler> = {
```
dlqqq, line 78: Can we merge the definition of `telemetryHandler` and `telemetry` so we can describe both of them in a single doc comment? Below is a rough sketch of my recommendation:

```
// The default telemetry plugin, which does nothing with obtained telemetry. Third-party extensions can override this to collect telemetry data from Jupyter Scheduler by disabling this plugin and exporting their own.
const telemetry: JupyterFrontEndPlugin<Scheduler.TelemetryHandler> = {
  ...
  activate: () => {
    return () => { /* noop */ }
  }
}
```

Open-source users are much more privacy and security conscious, so the word "telemetry" on its own may spook some users. A doc comment to clarify that "this does nothing by default" is good here.

src/index.tsx
```diff
@@ -223,13 +236,19 @@ async function activatePlugin(
   // Commands
 
   commands.addCommand(CommandIDs.showNotebookJobs, {
-    execute: async args => showJobsPanel(args as IJobsModel),
+    execute: async args => {
+      if (args['launcher']) {
+        eventLogger('launcher.show-jobs');
+      }
```
dlqqq, line 242: I'm a bit confused about the type of `args` here. We are casting it to `IJobsModel` below, and yet, the launcher command seems to only specify `args` to be `{ launcher: true }`. No action needed here, but this seems suspicious and a good place for bugs to hide.

dlqqq (changes requested):

src/index.tsx
3coins: Fixed this.

src/index.tsx
3coins: You should have access to the fork now.

dlqqq: Done in https://github.com/3coins/jupyter-scheduler/pull/3

dlqqq (approved):
Looks good! The doc comment was forgotten, but now that the implementation is obviously a no-op, I think the intent is sufficiently clear.

PR #150, opened 2022-10-13
add jobs table in job definition detail view

dlqqq:
- adds (jobSize?: number) prop to `ListJobsTable` to control page size of the table
- adds `ListJobsTable` table to `JobDefinition` with page size of 5 (please see preview below as well as previews of alternative page sizes 3 and 10)

Closes #134.

Preview:
page_size_5

Alternative page size 3:
page_size_3

Alternative page size 10:
job_size_10

PR #461, opened 2023-11-01
Removes zero-height style rule, which broke table display in Safari

### LINKED ISSUE #449: On Safari browser, the jobs section for in JobDefintion page is not visible
## Reproduce

1. Create a scheduled job, say, run a notebook untitled.ipynb every 2 minutes
2. After a few minutes when jobs are run, go to Job Definition page
3. On Chrome browser, you will see jobs that were started by this definition. On Safari, there is no jobs loaded. Both browsers are pointing to the same jupyterlab http://localhost:8888/lab

JasonWeill:
Fixes #449, opened by @funnypenguine. Removes a zero-height style rule on the job run list on the job definitions detail page. This style rule was introduced in #150, coauthored by @dlqqq and @andrii-i .

In Safari 17.0, I now see the job list view. See screen shot below. No regressions observed in the newest versions of Firefox ESR or Chrome.

image

### DIFF
--- a/src/components/advanced-table/advanced-table.tsx
+++ b/src/components/advanced-table/advanced-table.tsx
@@ -216,9 +216,7 @@ export function AdvancedTable<
   }
 
   const tableDiv = (
-    <div
-      style={height === 'auto' ? { flex: 1, height: 0 } : { maxHeight: height }}
-    >
+    <div style={height === 'auto' ? { flex: 1 } : { maxHeight: height }}>
       <TableContainer
         component={Paper}
         sx={{


andrii-i (changes requested):
Testing this on Safari Version 17.1 (18616.2.9.11.9, 18616) and getting a problem that `height: 0` hack was fixing: in Job Definition detail view, table with jobs is not rendered at all or is rendered empty.

To reproduce on my end:
1. Create a job definition that would execute every minute
2. After at least one job is created, see Job Definition detail view

In Chrome / how it's supposed to look:
image

How it looks for me in Safari:
image

andrii-i (approved):
Looks good, no regression. 

As advised by @dlqqq, this was introduce to deal with this issue: https://github.com/philipwalton/flexbugs/issues/197#issuecomment-378908438. I can’t reproduce the problem anymore. 100% height now works well within a flex item in a flex-item child in both Chrome and Safari. Example referenced in the thread now works correctly too. Probably browsers fixed the behavior https://jsfiddle.net/mLkp7L1j/125/

andrii-i:
@meeseeksdev please backport to 1.x

andrii-i:
@MeeseeksDev Hello

andrii-i:
@meeseeksdev Hello

PR #463, opened 2023-11-13
Pydantic v1 and v2 compatibility, add `pydantic_v1` module

### LINKED ISSUE #390: Not compatible with Pydantic 2

JasonWeill:
Fixes #390. Inspired by LangChain, adds a module `jupyter_scheduler.pydantic_v1` to expose the Pydantic v1 API to classes whether v1 or v2 is installed.

Updates dependencies to allow v1 and v2 to be used. Updates classes to use the new dependency.

Tested locally with both Pydantic v1 and v2; working in both cases.

### DIFF
--- a/jupyter_scheduler/pydantic_v1/__init__.py
+++ b/jupyter_scheduler/pydantic_v1/__init__.py
@@ -1,0 +1,25 @@
+from importlib import metadata
+
+## Create namespaces for pydantic v1 and v2.
+# Adapted from LangChain's langchain.pydantic_v1 module.
+#
+# This code must stay at the top of the file before other modules may
+# attempt to import pydantic since it adds pydantic_v1 and pydantic_v2 to sys.modules.
+#
+# This hack is done for the following reasons:
+# * Code will attempt to remain compatible with both pydantic v1 and v2 since
+#   both dependencies and dependents may be stuck on either version of v1 or v2.
+# * Creating namespaces for pydantic v1 and v2 should allow us to write code that
+#   unambiguously uses either v1 or v2 API.
+# * This change is easier to roll out and roll back.
+
+try:
+    from pydantic.v1 import *
+except ImportError:
+    from pydantic import *
+
+
+try:
+    _PYDANTIC_MAJOR_VERSION: int = int(metadata.version("pydantic").split(".")[0])
+except metadata.PackageNotFoundError:
+    _PYDANTIC_MAJOR_VERSION = 0
--- a/jupyter_scheduler/pydantic_v1/__init__.py
+++ b/jupyter_scheduler/pydantic_v1/__init__.py
@@ -6,9 +6,3 @@ try:
     from pydantic.v1 import *
 except ImportError:
     from pydantic import *
-
-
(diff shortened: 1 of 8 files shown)


jupyter_scheduler/pydantic_v1/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+try:
+    from pydantic.v1 import *
+except ImportError:
+    from pydantic import *
+
+
+try:
+    _PYDANTIC_MAJOR_VERSION: int = int(metadata.version("pydantic").split(".")[0])
+except metadata.PackageNotFoundError:
+    _PYDANTIC_MAJOR_VERSION = 0
```
dlqqq, line 25: This can be deleted, as setting this `_PYDANTIC_MAJOR_VERSION` does nothing except expose this information to LangChain's test suite: https://github.com/search?q=repo%3Alangchain-ai/langchain%20_PYDANTIC_MAJOR_VERSION&type=code

jupyter_scheduler/pydantic_v1/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# This code must stay at the top of the file before other modules may
+# attempt to import pydantic since it adds pydantic_v1 and pydantic_v2 to sys.modules.
+#
+# This hack is done for the following reasons:
+# * Code will attempt to remain compatible with both pydantic v1 and v2 since
+#   both dependencies and dependents may be stuck on either version of v1 or v2.
+# * Creating namespaces for pydantic v1 and v2 should allow us to write code that
+#   unambiguously uses either v1 or v2 API.
+# * This change is easier to roll out and roll back.
```
dlqqq, line 14: The code block is fairly obvious in purpose and not specific to LangChain. Therefore the comment should be shortened drastically:

```suggestion
# expose Pydantic v1 API, regardless of Pydantic version in current env
```

jupyter_scheduler/pydantic_v1/main.py
```diff
@@ -0,0 +1,4 @@
+try:
+    from pydantic.v1.main import *
+except ImportError:
+    from pydantic.main import *
```
dlqqq, line 4: (not specific to this file, just need a target per review comment)

We should add `dataclasses.py` in case it is needed in the future. See LangChain's `pydantic_v1` module source for reference: https://github.com/langchain-ai/langchain/tree/master/libs/langchain/langchain/pydantic_v1

dlqqq (changes requested):

dlqqq (approved):
Awesome work!

PR #472, opened 2024-01-12
Emit telemetry events on success and failure of the Create Job, Create Job Definition, Create Job from Job Definition hooks

### LINKED ISSUE #471: Emit telemetry events on create job forms
### Problem
Would like to have telemetry data emitted for `Compute type` dropdown on create job page, and also for success / failure events after clicking on create job button. 

### Proposed Solution
1. When a compute type is selected using the dropdown, the click event along with the selected compute type will be logged using the `useEventLogger` hook.
2. In the case of successful job creation or job failure, the success/failure event is logged using the `useEventLogger` hook.

andrii-i:
Emit telemetry events on success and failure of the Create Job, Create Job Definition, Create Job from Job Definition hooks. 
* Log object's `body.name` string is now following a new convention: `org.jupyter.jupyter-scheduler.SOURCE.EVENT[.STATUS]` (vs old convention `org.jupyter.jupyter-scheduler.SOURCE.EVENT`) where `STATUS` is either `"success"` or `"failure"`
* `CreateJob` now emits `"create-job"` or `"create-job-definition"` for `EVENT` instead of just `"create"` depending on if job or job definition is being created
* Log object's `body` now has optional `detail` parameter storing error message (or other string with event detail)
```ts
type EventLog = {
  body: { name: string; detail?: string };
  timestamp: Date;
};
```
* Improve error catching logic by catching errors as `unknown` type and using type guarding instead of assuming all errors caught would be of `Error` type (ts runtime error type) which is not guaranteed.
* Fix #471 

Emitted objects examples:
Screenshot 2024-01-18 at 2 52 01 PM

Screenshot 2024-01-18 at 2 50 31 PM

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,10 +1,12 @@
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import React from 'react';
 
-export type Logger = (eventName: string) => void;
-export const LogContext = React.createContext<Logger>((eventName: string) => {
-  /*noop*/
-});
+export type Logger = (eventName: string, error?: Error) => void;
+export const LogContext = React.createContext<Logger>(
+  (eventName: string, error?: Error) => {
+    /*noop*/
+  }
+);
 
 // Context to be overridden with JupyterLab context
 const TranslatorContext = React.createContext<ITranslator>(nullTranslator);
--- a/src/context.ts
+++ b/src/context.ts
@@ -3,7 +3,7 @@ import React from 'react';
 
 export type Logger = (eventName: string, error?: Error) => void;
 export const LogContext = React.createContext<Logger>(
-  (eventName: string, error?: Error) => {
+  (eventName: string, eventDetail?: string) => {
     /*noop*/
   }
 );
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,7 +1,7 @@
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import React from 'react';
 
-export type Logger = (eventName: string, error?: Error) => void;
+export type Logger = (eventName: string, eventDetail?: string) => void;
 export const LogContext = React.createContext<Logger>(
   (eventName: string, eventDetail?: string) => {
     /*noop*/
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -42,7 +42,7 @@ export const NotebookJobsPanelId = 'notebook-jobs-panel';
 export { Scheduler } from './tokens';
 
 type EventLog = {
-  body: { name: string; detail?: string } };
+  body: { name: string; detail?: string };
   timestamp: Date;
 };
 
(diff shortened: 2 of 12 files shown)


src/mainviews/create-job-from-definition.tsx
```diff
@@ -139,6 +139,7 @@ export function CreateJobFromDefinition(
         createJobFromDefinitionModel
       )
       .then(response => {
+        log('create-job-from-definition.create-job.success');
```
bhadrip, line 142: can we log the error response - http response code and message ?

andrii-i: Added optional `error` parameter with error message and HTTP status code to log object's `body` : https://github.com/jupyter-server/jupyter-scheduler/pull/472/files#diff-0b5adbfe7b36e4ae2f479291e20152e33e940f7f265162d77f40f6bdb5da7405R44-R47

src/index.tsx
```diff
@@ -41,6 +41,11 @@ export namespace CommandIDs {
 export const NotebookJobsPanelId = 'notebook-jobs-panel';
 export { Scheduler } from './tokens';
 
+type EventLog = {
+  body: { name: string; error?: { message?: string; httpStatusCode?: number } };
```
3coins, line 45: Capturing the error detail/string should be sufficient to capture the error.
```suggestion
  body: { name: string; detail?: string } };
```

src/context.ts
```diff
@@ -1,10 +1,12 @@
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import React from 'react';
 
-export type Logger = (eventName: string) => void;
-export const LogContext = React.createContext<Logger>((eventName: string) => {
-  /*noop*/
-});
+export type Logger = (eventName: string, error?: Error) => void;
+export const LogContext = React.createContext<Logger>(
+  (eventName: string, error?: Error) => {
```
3coins, line 6:
```suggestion
  (eventName: string, eventDetail?: string) => {
```

src/context.ts
```diff
@@ -1,10 +1,12 @@
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import React from 'react';
 
-export type Logger = (eventName: string) => void;
-export const LogContext = React.createContext<Logger>((eventName: string) => {
-  /*noop*/
-});
+export type Logger = (eventName: string, error?: Error) => void;
```
3coins, line 4: Making error as `Error` type is too specific to HTTP errors. Though the intent here is to primarily capture the HTTP errors, having something simpler can accomplish this. 

```suggestion
export type Logger = (eventName: string, eventDetail?: string) => void;
```

andrii-i:
@3coins thank you for looking into this, implemented your comments re not logging an `httpResponseCode`. Also added improved error catching logic as discussed.

JasonWeill (commented):
"Unknown error occurred" is a generic message, but it may not inspire confidence. Do we have any idea about what sorts of errors might trigger the generic message? If so, can we provide any additional guidance? Something like, "An error occurred. Please try again."?

andrii-i:
@JasonWeill thank you for looking into this. Any error / error object that does not have fields expected from [es5 Error object](https://github.com/microsoft/TypeScript/blob/1982349339b9c1fd78c93195d572ffc6fd5041b0/src/lib/es5.d.ts#L1057) would trigger the generic error message. Most probably to occur are connection failures, connection timeouts, third-party libraries throwing something else than es5 `Error` (other objects or just strings). So the message needs to be generic, I think the message that you proposed works well, updated the function to use it.

JasonWeill (approved):
Looks good, thanks!

3coins (approved):

PR #480, opened 2024-02-09
Make server extension verification call during extension startup non-blocking

### LINKED ISSUE #479: Lab extension makes a blocking call to the server extension on startup
## Description

The front-end scheduler extension makes a blocking network call to the server extension in its start-up `activatePlugin` function:

https://github.com/jupyter-server/jupyter-scheduler/blob/1b50f81f966c9e0146f91b1606fd78519857527e/src/index.tsx#L154-L156

This can cause a delay in JupyterLab startup, even if the user does not intend to use the scheduler extension.

## Expected behavior

The scheduler should make asynchronous calls during its `activatePlugin` function.

andrii-i:
* Make server extension verification call in index.ts non-blocking, refactor it into a separate function
* Fixes #479 

To test:
1. Add a sleep statement like `await asyncio.sleep(20)` (20 seconds) to [JobHandler.get ](https://github.com/jupyter-server/jupyter-scheduler/blob/1b50f81f966c9e0146f91b1606fd78519857527e/jupyter_scheduler/handlers.py#L189) as so:

```python
class JobHandler(ExtensionHandlerMixin, JobHandlersMixin, APIHandler):
    @authenticated
    async def get(self, job_id=None):
        await asyncio.sleep(20)
        ...
```
2. Launch the extension with and without this PR, note that with this PR startup is not blocked

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@jupyterlab/coreutils": "^6",
     "@jupyterlab/filebrowser": "^4",
     "@jupyterlab/launcher": "^4",
-    "@jupyterlab/rendermime-interfaces": "^3.9.0",
+    "@jupyterlab/rendermime-interfaces": "^3.8.0",
     "@jupyterlab/services": "^7",
     "@jupyterlab/translation": "^4",
     "@jupyterlab/ui-components": "^4",
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "@jupyterlab/coreutils": "^6",
     "@jupyterlab/filebrowser": "^4",
     "@jupyterlab/launcher": "^4",
+    "@jupyterlab/rendermime-interfaces": "^3.9.0",
     "@jupyterlab/services": "^7",
     "@jupyterlab/translation": "^4",
     "@jupyterlab/ui-components": "^4",
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -47,11 +47,6 @@ type EventLog = {
   timestamp: Date;
 };
 
-type verifyServerExtensiontProps = {
-  api: SchedulerService;
-  trans: IRenderMime.TranslationBundle;
-};
-
 /**
  * Call API to verify that the server extension is actually installed.
  */
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -12,6 +12,7 @@ import {
 import { FileBrowser, IFileBrowserFactory } from '@jupyterlab/filebrowser';
 import { ILauncher } from '@jupyterlab/launcher';
 import { INotebookTracker } from '@jupyterlab/notebook';
+import { IRenderMime } from '@jupyterlab/rendermime-interfaces';
 import { Contents, ServerConnection } from '@jupyterlab/services';
 import { ITranslator } from '@jupyterlab/translation';
 
@@ -45,6 +46,33 @@ type EventLog = {
   body: { name: string; detail?: string };
   timestamp: Date;
 };
+
+type verifyServExtProps = {
+  api: SchedulerService;
+  trans: IRenderMime.TranslationBundle;
+};
+
+/**
+ * Call API to verify that the server extension is actually installed.
+ */
+async function verifyServExt(props: verifyServExtProps) {
+  try {
+    await props.api.getJobs({ max_items: 0 });
+  } catch (e: unknown) {
+    // in case of 404, show missing server extension dialog and return
+    if (
+      e instanceof ServerConnection.ResponseError &&
+      e.response.status === 404
+    ) {
+      showDialog({
+        title: props.trans.__('Jupyter Scheduler server extension not found'),
+        body: SERVER_EXTENSION_404_JSX,
+        buttons: [Dialog.okButton()]
+      }).catch(console.warn);
+      return;
+    }
+  }
+}
 
 /**
  * Initialization data for the jupyterlab-scheduler extension.
@@ -150,24 +178,7 @@ async function activatePlugin(
 ): Promise<void> {
   const trans = translator.load('jupyterlab');
   const api = new SchedulerService({});
-
-  // Call API to verify that the server extension is actually installed
-  try {
-    await api.getJobs({ max_items: 0 });
-  } catch (e: unknown) {
-    // in case of 404, show missing server extension dialog and return
-    if (
-      e instanceof ServerConnection.ResponseError &&
-      e.response.status === 404
-    ) {
-      showDialog({
-        title: trans.__('Jupyter Scheduler server extension not found'),
-        body: SERVER_EXTENSION_404_JSX,
-        buttons: [Dialog.okButton()]
-      }).catch(console.warn);
-      return;
-    }
-  }
+  verifyServExt({ api, trans });
 
   const { commands } = app;
   const fileBrowserTracker = browserFactory.tracker;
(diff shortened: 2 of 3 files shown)


src/index.tsx
```diff
@@ -46,6 +47,33 @@ type EventLog = {
   timestamp: Date;
 };
 
+type verifyServExtProps = {
```
JasonWeill, line 50: Can you please spell out `Server` and `Extension`? Space isn't super tight here, and this helps with accessibility, in case someone is using a screen reader or translation software to read out code.

src/index.tsx
```diff
@@ -46,6 +47,33 @@ type EventLog = {
...
+  api: SchedulerService;
+  trans: IRenderMime.TranslationBundle;
+};
+
+/**
+ * Call API to verify that the server extension is actually installed.
+ */
+async function verifyServExt(props: verifyServExtProps) {
```
JasonWeill, line 58: Please expand `Server` and `Extension`, as above

src/index.tsx
andrii-i: That's an interesting point about accessibility, have not considered it before. Will implement the suggestion, will look more into negative impacts of using abbreviations in code as I quite like using them.

src/index.tsx
```diff
@@ -46,6 +47,33 @@ type EventLog = {
...
+  try {
+    await props.api.getJobs({ max_items: 0 });
```
bhadrip, line 60: How did you test this change ?  What is the behavior of `activatePlugin`  ? in my understanding, until this function completes, Jupyterlab is going to show the loading screen. 

It is not clear how sync vs async helps in solving this issue.

src/index.tsx
```diff
@@ -46,6 +47,33 @@ type EventLog = {
...
 
+type verifyServerExtensiontProps = {
+  api: SchedulerService;
+  trans: IRenderMime.TranslationBundle;
+};
+
+/**
+ * Call API to verify that the server extension is actually installed.
+ */
+async function verifyServerExtension(props: verifyServerExtensiontProps) {
```
dlqqq, line 58: Using an inline type definition is preferable for simple, local types (i.e. ones that are not exported):

```suggestion
async function verifyServerExtension(props: {
  api: SchedulerService;
  trans: IRenderMime.TranslationBundle;
}) {
```

This syntax is significantly shorter and allows developers to easily see the type of `props` without having to look up a type definition above.

I'm leaving this up to you. However, if you want to keep the named type, you should correct the spelling and capitalize it per convention.

src/index.tsx
```diff
@@ -12,6 +12,7 @@ import {
 import { FileBrowser, IFileBrowserFactory } from '@jupyterlab/filebrowser';
 import { ILauncher } from '@jupyterlab/launcher';
 import { INotebookTracker } from '@jupyterlab/notebook';
+import { IRenderMime } from '@jupyterlab/rendermime-interfaces';
```
dlqqq, line 15: `@jupyterlab/rendermime-interfaces` needs to be declared in `package.json`. Here you want to set its version range spec to `^4`, but when backporting, you should edit it to `^3`.

src/index.tsx
andrii-i: Note that we are not awaiting [`verifyServerExtension` call inside `activatePlugin`](https://github.com/jupyter-server/jupyter-scheduler/pull/480/files#diff-0b5adbfe7b36e4ae2f479291e20152e33e940f7f265162d77f40f6bdb5da7405R181) vs having blocking `await api.getJobs({ max_items: 0 });` call there before. When `async` function is called without `await`, the code execution continues to the next line without waiting for the promise returned by the async function to resolve or reject, the main execution flow is not blocked, the asynchronous operation runs in the background and resolves whenever it finishes independent of the main execution flow.  

As mentioned in the PR description, to test:

1. Add a sleep statement like `await asyncio.sleep(20)` (20 seconds) to [JobHandler.get ](https://github.com/jupyter-server/jupyter-scheduler/blob/1b50f81f966c9e0146f91b1606fd78519857527e/jupyter_scheduler/handlers.py#L189) as so:

```python
class JobHandler(ExtensionHandlerMixin, JobHandlersMixin, APIHandler):
    @authenticated
    async def get(self, job_id=None):
        await asyncio.sleep(20)
        ...
```
2. Launch the extension with and without this PR, note that with this PR startup is not blocked

andrii-i: @bhadrip please let me know if this answers your questions and resolves your concerns.

src/index.tsx
```diff
@@ -150,24 +178,7 @@ async function activatePlugin(
 ): Promise<void> {
```
dlqqq, line 178: Because you've removed the only blocking `await` statement, you can actually just edit the definition `activatePlugin` to be entirely synchronous. Change its definition to:

```ts
function activatePlugin(
  app: JupyterFrontEnd,
  browserFactory: IFileBrowserFactory,
  notebookTracker: INotebookTracker,
  translator: ITranslator,
  restorer: ILayoutRestorer,
  advancedOptions: Scheduler.IAdvancedOptions,
  telemetryHandler: Scheduler.TelemetryHandler,
  launcher: ILauncher | null
): void {
  ...
}
```

src/index.tsx
dlqqq: The JupyterLab splash shows while any extensions are still activating. Removing the `await` statement from `activatePlugin()` will allow the function to complete regardless of how long `api.getJobs()` takes, meaning that Jupyter Scheduler will no longer be forcing the JupyterLab splash / loading screen to be shown after this PR.

While I agree testing would be nice here, testing should not block this PR, as the expected behavior is immediately apparent from code inspection.

dlqqq (changes requested):

src/index.tsx
bhadrip: makes sense, thanks for the explanation.

src/index.tsx
andrii-i: Added as `"@jupyterlab/rendermime-interfaces": "^3.9.0"` as this is the latest version of the package released, version [used in latest jupyterlab](https://github.com/jupyterlab/jupyterlab/blob/49f6ec28b738243a53a947d544bebafafcbd4a73/dev_mode/package.json#L97).

src/index.tsx
andrii-i: That's a really good point. Done.

src/index.tsx
andrii-i: Scheduler depends on JupyterLab 4.0.0 or above. JupyterLab 4.0 depends on @jupyterlab/rendermime-interfaces ~3.8.0  (any 3.8.x) and JupyterLab 4.1 depends on @jupyterlab/rendermime-interfaces ~3.9.0  (any 3.9.x). Based on this, pinned @jupyterlab/rendermime-interfaces as 3.8.0 and above (^3.8.0) to be able to support any JupyterLab 4 version released so far.

dlqqq (approved):
👍

package.json
```diff
@@ -62,6 +62,7 @@
     "@jupyterlab/coreutils": "^6",
     "@jupyterlab/filebrowser": "^4",
     "@jupyterlab/launcher": "^4",
+    "@jupyterlab/rendermime-interfaces": "^3.8.0",
```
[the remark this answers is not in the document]
andrii-i: Thank you @dlqqq

PR #481, opened 2024-02-10
[1.x] Make server extension verification call during extension startup non-blocking (#480)

andrii-i:
* Backports #480


yarn.lock
```diff
@@ -1982,6 +1982,14 @@
...
     resize-observer-polyfill "^1.5.1"
 
+"@jupyterlab/rendermime-interfaces@^3.0.0":
+  version "3.9.0"
+  resolved "https://registry.yarnpkg.com/@jupyterlab/rendermime-interfaces/-/rendermime-interfaces-3.9.0.tgz#9f4a54a20ca96954cffb419617d6da2957c3b26f"
+  integrity sha512-bjaqMkWXkWbtUmnJng5L813p3sSAAHmyj02DRppnCBgwT5C9l9JKVG7/NRaVYpbL5CtZBuR9oMpHKbmUP695pg==
+  dependencies:
+    "@lumino/coreutils" "^1.11.0 || ^2.1.2"
+    "@lumino/widgets" "^1.37.2 || ^2.3.1"
+
```
dlqqq, line 1992: The JL3 and JL4 versions of the `@jupyterlab/rendermime-interfaces` package both share the same major version. `yarn` is resolving this to v3.9.0 in our lockfile, the same version being used in the `main` branch now. While the lack of a major version bump indicates that this should be safe for users, we don't want 2 different major versions of Lumino in our dependencies while developing and building. This likely is causing the `tsc` build errors that you faced; notice that they all come from the new Lumino 2 package being pulled in.

To fix this, you'll want to use [`yarn` selective dependency resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) and make `"@jupyterlab/rendermime-interfaces"` be resolved to `"3.6.7"`, [the latest JL3 version of this package](https://github.com/jupyterlab/jupyterlab/blob/v3.6.7/packages/rendermime-interfaces/package.json).

Then, after you see that the Lumino 2 packages are no longer present in the lockfile, try reverting the changes to `tsconfig.json` and see if the build still passes without any changes to that file. 👍

dlqqq (changes requested):

yarn.lock
andrii-i: Specifying `"@jupyterlab/rendermime-interfaces": "3.6.7"` in `package.json "resolutions"` does not prevent a dependency on `"@jupyterlab/rendermime-interfaces": "3.9.0"` being pulled in.

While working on the backport, noticed incorrect usage of the translator in the main PR, opened PR #483 to fix it. This would also remove the need to depend on @jupyterlab/rendermime-interfaces .  Please give it a look when possible.
Then I’ll backport this PR and #480 together as this PR singificantly simplifies the dependency graph for a backport and prevents lumino v2 from being pulled into dependencies.

andrii-i: Note how #483 [yarn.lock changes](https://github.com/jupyter-server/jupyter-scheduler/pull/483/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de) eliminate problematic lumino v2 transitory dependencies.

andrii-i: Lumino problem is no more, figured it out (had to pin all lumino packages that could be v1 or v2 as v1 in resolutions) but am still getting mui errors.

For mui errors, had to resort to `skipLibCheck: true` as you advised. Otherwise I’m getting mui type errors below. I tried everything I could think of but to no avail: updating ts version, updating react and/or react types version, upgrading / downgrading mui/base and mui/types versions. As evident from the error text errors are due to a type mismatch. Based on my research, this should be due to mismatch between react / react-types / mui / mui-types versions. But they don’t specify each other as dependencies so it’s not clear which versions should be compatible…

```
      node_modules/@mui/base/Unstable_Popup/Popup.d.ts(13,70): error TS2707: Generic type 'ElementType' requires between 0 and 1 type arguments.
      node_modules/@mui/material/Popper/Popper.d.ts(46,17): error TS2707: Generic type 'ElementType' requires between 0 and 1 type arguments.
      node_modules/@mui/material/Popper/Popper.d.ts(53,16): error TS2707: Generic type 'ElementType' requires between 0 and 1 type arguments.
      node_modules/@mui/system/cssVars/cssVarsParser.d.ts(2,6): error TS1337: An index signature parameter type cannot be a union type. Consider using a mapped object type instead.
```

andrii-i:
Updated the PR description based on the current state of the PR

yarn.lock
dlqqq: Yes, I researched this and the need for `skipLibCheck: true` seems unavoidable. I checked out the MUI source repo at our resolved version, and they are using TypeScript v5 while JupyterLab v3 uses TypeScript v4.

dlqqq:
@andrii-i I've fixed a few issues with our `package.json` and simplified it. Please check it out: https://github.com/andrii-i/jupyter-scheduler/pull/1

dlqqq:
The failing command in the "Build / build" workflow is:

```
jupyter server extension list 2>&1 | grep -ie "jupyter_scheduler.*OK"
```

Looking at the logs, this is because of an unrelated error with the default JL3 installation:

```
+ jupyter labextension list
      jupyter_scheduler 1.8.0 OK
JupyterLab v3.6.7
/opt/hostedtoolcache/Python/3.11.8/x64/share/jupyter/labextensions
        jupyterlab_pygments v0.3.0 enabled  X (python, jupyterlab_pygments)
        @jupyterlab/scheduler v1.8.0 enabled  X


   The following extensions are outdated:
        @jupyterlab/scheduler
        
   Consider checking if an update is available for these packages.


"jupyterlab_pygments@0.3.0" is not compatible with the current JupyterLab
Conflicting Dependencies:
JupyterLab              Extension      Package
>=3.6.7 <3.7.0          >=4.0.8 <5.0.0 @jupyterlab/application
```

This is because of https://github.com/jupyterlab/jupyterlab/issues/14351. Basically, we need to always make sure our version ranges overlap with the ones declared in `staging/package.json` in the JupyterLab source. Since this is my error, I'm opening a PR on your fork to fix this.

dlqqq (approved):
👍

PR #483, opened 2024-02-12
Fix translator usage, remove @jupyterlab/rendermime-interfaces dependency

andrii-i:
PR #480 introduced [incorrect usage of `translator`](https://github.com/jupyter-server/jupyter-scheduler/blob/d17ed34897448ef28e1976c37ff25950429daf83/src/index.tsx#L53-L66): `IRenderMime.TranslationBundle` is passed directly to `verifyServerExtension` function. Correct pattern is passing `ITranslator` to any component and then instantiating `IRenderMime.TranslationBundle` via `translator.load('jupyterlab');` as it's done everywhere in the codebase, for example [here](https://github.com/jupyter-server/jupyter-scheduler/blob/d17ed34897448ef28e1976c37ff25950429daf83/src/notebook-jobs-panel.tsx#L69). 

As a result of this change, translator inside `verifyServerExtension` function works correctly and explicit dependency on `@jupyterlab/rendermime-interfaces ` is not needed (this will also simplify PR  #481 that backports #480).

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "@jupyterlab/coreutils": "^6",
     "@jupyterlab/filebrowser": "^4",
     "@jupyterlab/launcher": "^4",
-    "@jupyterlab/rendermime-interfaces": "^3.8.0",
     "@jupyterlab/services": "^7",
     "@jupyterlab/translation": "^4",
     "@jupyterlab/ui-components": "^4",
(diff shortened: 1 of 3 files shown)


andrii-i:
Looking at [yarn.lock changes](https://github.com/jupyter-server/jupyter-scheduler/pull/483/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de), it is evident that this PR eliminates problematic lumino v2 transitory dependencies.

dlqqq (approved):

PR #486, opened 2024-02-16
[1.1.x] Make server extension verification call during extension startup non-blocking (#480)

andrii-i:
* Backports #480


andrii-i:
Readthedocs failure is expected as we don't want to build docs for this branch, don't have config files such as `.readthedocs.yml`, `docs/conf.py`. Looking into adjusting readthedocs website settings to not build docs for this branch by marking it as inactive (https://docs.readthedocs.io/en/stable/versions.html#active).

PR #500, opened 2024-03-14
Add an option to package input folder; auto-download files from the staging directory to output

### LINKED ISSUE #407: Support for packaging dependency files
### Problem

Currently there is no way to submit dependencies for a job run

### Proposed Solution

Provide ability to include multiple input files for a single job, but only one file that gets "run". Another way of phrasing this: specify both a directory and some "main" file as an input file.

andrii-i:
When working with notebooks, users frequently import and use additional files such as datasets, images, and scripts inside their notebook's cells. Providing support for packaging such files would ensure that notebooks can have all essential resources available when executed as a job. This would make Jupyter Scheduler more flexible and able to accommodate more types or workflows providing better value to people who use it.

This PR adds an option to package input folder (folder where input notebook is located) and all nested files and sub-folders within it during the job or job definition creation. It also makes so output files are automatically downloaded to output folder when job runs, not on download button click like before. 

Also adds `DownloadRunner` and `DownloadManager` components that manage downloads in asynchronous environment via multiprocess-safe queue persisted through database table.

When package input folder option is active:
* All files and sub-folders within input folder are copied to the staging area together with the snapshot of an input notebook
* Original organization of files and folders (file tree) is maintained, input notebook is able to access them via relative file paths when it is ran as a job
* When job is ran, the contents of the job's staging directory (snapshot of the input notebook, packaged input files, output files, and side effects) are copied into the output folder. User can access the output folder by clicking "Files" shortcut under "Output files" in the list view (it will open the folder in JupyterLab filebrowser).
* For jobs, packaged input files and side effects are tracked in `Job.packaged_files`, if any of them is deleted from output folder, user gets an option to re-downloaded them via UI (matches existing behavior for snapshot of the input notebook and output files)
* For job definitions, packaged input files and side effects are tracked in `JobDefinition.packaged_files`
* [Modifies notebook / job execution logic](https://github.com/jupyter-server/jupyter-scheduler/pull/500/commits/b124105928634f426e58f740ea6dc99f5078adf3) so that kernel executing the notebook always operates in the same directory as the notebook being executed (`cwd` parameter of the `ExecutePreprocessor`). Additionally pass intended path of execution context to preprocessor via metadata `{"metadata": {"path": notebook_dir}}` argument of the preprocessor call.
* Side effects of running an input notebook are added to `Job.packaged_files` and copied to the output folder together with other files

Fixes #407

Before:
image
image

After:
Screenshot 2024-04-23 at 12 51 33 AM

Screenshot 2024-04-02 at 1 46 20 PM

Re-download option when any of the files is deleted from the output folder:
Screenshot 2024-04-02 at 1 47 36 PM

image

Screenshot 2024-04-23 at 1 43 49 AM


bhadrip:
Can we support the following use case:

```
jupyter-root-dir
  dep-dir
     init.py
  project1-dir
     analysis.ipynb (depends on dep-dir/init.py)
  project2-dir
     report-generation.ipynb (depends on dep-dir/init.py)
```
2 separate notebook jobs could depend on the same python file:
Job1 - `analysis.ipynb`
Job2 - `report-generation.ipynb`

With the current approach the customer has to copy the `dep-dir` to every project folder.

JasonWeill:
Upon first view, the term "Package input folder" doesn't mean anything to me. I don't know why, as a user, I would want to do that, or whether I would need to. This change also doesn't modify the documentation at all.

I would recommend:

1. Modifying the documentation to describe what "package input folder" does
2. Adding some kind of "info" ℹ️ button or expander near the new checkbox to explain what the option does, briefly

src/components/files-directory-link.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export function FilesDirectoryLink(props: {
+  jobFile: Scheduler.IJobFile;
+  app: JupyterFrontEnd;
+}): JSX.Element | null {
+  const trans = useTranslator('jupyterlab');
+  return (
+    <Link
+      href={`/lab/tree/${props.jobFile.file_path}`}
+      title={trans.__('Open output directory with files')}
```
JasonWeill, line 15: This title doesn't seem clear enough. Would "Open output directory in file browser" be clearer, to emphasize that clicking this link would cause a change in the file browser, not in the main work area?

src/mainviews/detail-view/job-definition.tsx
```diff
@@ -230,6 +230,12 @@ export function JobDefinition(props: IJobDefinitionProps): JSX.Element {
         value: model.timezone ?? '',
         label: trans.__('Time zone')
       }
+    ],
+    [
+      {
+        value: model.packageInputFolder ? trans.__('Yes') : trans.__('No'),
+        label: trans.__('Package input folder')
```
JasonWeill, line 237: This is not consistent with the other headers that refer to a verb; would "Input folder packaged" be better?

src/mainviews/detail-view/job-detail.tsx
```diff
@@ -249,6 +249,12 @@ export function JobDetail(props: IJobDetailProps): JSX.Element {
       {
         value: timestampLocalize(props.model.endTime ?? ''),
         label: trans.__('End time')
+      },
+      {
+        value: props.model.packageInputFolder
+          ? trans.__('Yes')
+          : trans.__('No'),
+        label: trans.__('Package input folder')
```
JasonWeill, line 257: This is not consistent with the other headers that refer to a verb; would "Input folder packaged" be better?

andrii-i:
Kicking CI

dlqqq (changes requested):
Thanks Andrii! I left a few comments regarding the auto-download capability being introduced in this branch, but didn't complete a full code review.

My main concern with this PR is that 1) the auto-download capability being proposed here is a potential breaking change, and 2) the changes to the download logic are not required to implement the 'include parent dir' capability, which was the original intent of this PR.

Let's explore the possibility opening another PR with just the 'include parent dir' capability such that it can be reviewed, merged, and released separately.

sravyasdh:
> Can we support the following use case:
> 
> ```
> jupyter-root-dir
>   dep-dir
>      init.py
>   project1-dir
>      analysis.ipynb (depends on dep-dir/init.py)
>   project2-dir
>      report-generation.ipynb (depends on dep-dir/init.py)
> ```
> 
> 2 separate notebook jobs could depend on the same python file: Job1 - `analysis.ipynb` Job2 - `report-generation.ipynb`
> 
> With the current approach the customer has to copy the `dep-dir` to every project folder.

@andrii-i can you confirm if this usecase can be supported? Can we support changing the input folder path on the UI?  Rest of this lgtm

andrii-i:
@sravyasdh thank you for looking into this PR. In the initial version only packaging the input file's folder via checkbox control will be supported so use case in question will not be supported. Reason is that it's not clear if users would prefer more complex UI/UX that would allow packaging of the folder one level above or packaging of the arbitrary folder or ability to input folder path on the UI. So it makes sense to start with simpler implementation, let users try it and get feedback from them, and update the implementation based on feedback if needed.

andrii-i:
For package input folder, superseded by #510.
For auto-download, either this PR will be reopened or a new PR would be created based on relevant commits.

PR #496, opened 2024-03-11
Changed column header to "Input file" in Notebook Job Definitions

srdas:
The Notebook Job Descriptions table now shows "Input file" in the column header instead of "Input filename". 
This makes the column header consistent across the *Notebook Jobs* tab and the  *Notebook Job Descriptions* tab.

Fixes #492

PR #501, opened 2024-03-19
[1.x] Changed column header to "Input file" in Notebook Job Definitions (#496)

srdas:
* Changed column header to Input file in Notebook Job Definitions

* Update list-jobs.tsx for column header change in Notebook Jobs table

Changed the column headers for the inout file in both Notebook Jobs and Notebook Job Definitions to 'Input filename" - we choose “filename” over “file” is because we don’t show the absolute or relative path, just the name of the file in any directory.

* Updated E2E snapshots for test


andrii-i:
Thank you @srdas.

andrii-i:
Kicking CI

PR #510, opened 2024-04-24
Package input files (no autodownload, no multiprocessing DownloadManager)

andrii-i:
This PR adds an option to package input folder (folder where input notebook is located) and all nested files and sub-folders within it during the job or job definition creation. 

In terms of features, this is a subset of PR #500. This PR does not automatically download output files to output folder when job runs and therefore has no need need to schedule downloads from multiple processes and components that would manage it (`DownloadRunner` and `DownloadManager` from #500). Besides making this PR more focused in terms of functionality, this makes changes introduced by this PR non-breaking.

When package input folder option is active:
* All files and sub-folders within input folder are copied to the staging area together with the snapshot of an input notebook
* Original organization of files and folders (file tree) is maintained, input notebook is able to access them via relative file paths when it is ran as a job
* After downloading job files, users can access the output folder by clicking "Files" shortcut under "Output files" in the list view (it will open the folder in JupyterLab filebrowser).
* For jobs, packaged input files and side effects are tracked in `Job.packaged_files`, if any of them is deleted from output folder, user gets an option to re-downloaded them via UI (matches existing behavior for snapshot of the input notebook and output files)
* For job definitions, packaged input files and side effects are tracked in `JobDefinition.packaged_files`
* Modifies notebook / job execution logic so that kernel executing the notebook always operates in the same directory as the notebook being executed (`cwd` parameter of the `ExecutePreprocessor`). Additionally pass intended path of execution context to preprocessor via metadata `{"metadata": {"path": notebook_dir}}` argument of the preprocessor call.
* Side effects of running an input notebook are added to `Job.packaged_files` and copied to the output folder together with other files

Fixes https://github.com/jupyter-server/jupyter-scheduler/issues/407

Before:
image
image

After:
Screenshot 2024-04-25 at 11 31 38 AM

Screenshot 2024-04-25 at 11 31 54 AM

Screenshot 2024-04-02 at 1 46 20 PM

Re-download option when any of the files is deleted from the output folder:
Screenshot 2024-04-02 at 1 47 36 PM

image


andrii-i:
Kicking CI

andrii-i:
Kicking CI

JasonWeill (commented):
This is looking better! Could you please update the screenshots and text in the docs to reflect this new option? I can help with the copy editing.

src/components/input-folder-checkbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+  let helperText: string;
+  if (inputFilePath.length) {
+    const inputFolder = `/${inputFilePath.join('/')}`;
+    helperText = trans.__(
+      'The scheduled job will have access to all files under %1',
+      inputFolder
+    );
+  } else {
+    helperText = trans.__(
+      'The scheduled job will have access to all files under input file’s folder'
```
JasonWeill, line 28: The character in `file’s` is a closing single quote. It should be an apostrophe:

```suggestion
      'The scheduled job will have access to all files under input file\'s folder'
```

andrii-i:
Kicking CI

andrii-i:
> Could you please update the screenshots and text in the docs to reflect this new option? I can help with the copy editing.

@JasonWeill updated "Submit the Create Job form" section of the Users page with a screenshot and text mentioning "Run job with input folder" option ([code](https://github.com/jupyter-server/jupyter-scheduler/pull/510/files#diff-c58422732a1288eb7673d0e8ad5d81b7447a9523d1b2c7324c776da3948ecd01), [readthedocs preview](https://jupyter-scheduler--510.org.readthedocs.build/en/510/users/index.html#submit-the-create-job-form)).

I tried different options like adding an example "Use this to, for example, access data files or images from notebook's cells." But ultimately what I have in the PR now works well and matches level of detail given in the paragraphs around it, would be interested to learn your opinion on this.

docs/users/index.md
```diff
@@ -49,6 +49,8 @@ To create a _job_ or _job definition_ from an open notebook, click on a “Creat
 
 Give your notebook job or job definition a name, choose an environment to run it in, select its output formats, and provide parameters that are set as local variables when your notebook gets executed. This parameterized execution is similar to the one used in [Papermill](https://papermill.readthedocs.io/en/latest/).
 
+Select “Run job with input folder” to ensure the scheduled job will have access to all files within the same folder as the input file.
```
JasonWeill, line 52:
```suggestion
If you check "Run job with input folder", the scheduled job will have access to all files within the same folder as the input file. 
The scheduler will copy all files from the input file to a staging directory at runtime, and will make these files available for download after the job has finished.
Use caution with this option if your input file's directory has many large files in it.
```

This should indicate both the benefit and the risk of the new option.

src/components/input-folder-checkbox.tsx
```diff
@@ -0,0 +1,46 @@
...
+  let helperText: string;
+  if (inputFilePath.length) {
+    const inputFolder = `/${inputFilePath.join('/')}`;
+    helperText = trans.__(
+      'The scheduled job will have access to all files under %1',
+      inputFolder
+    );
+  } else {
+    helperText = trans.__(
+      "The scheduled job will have access to all files under input file's folder"
```
JasonWeill, line 28:
```suggestion
      "The scheduled job will have access to all files under the input file's folder"
```

Please also update the screen shot in the docs.

andrii-i:
Kicking CI

JasonWeill (approved):
Change looks good! Thanks for the great work on this.

PR #513, opened 2024-04-30
Add tests for Scheduler job and job definition creation with input folder, refactor execution manager test

andrii-i:
* Add tests for `Scheduler` job and job definition creation with input folder
* Refactor `DefaultExecutionManager` test to use db fixtures


conftest.py
```diff
@@ -13,6 +13,8 @@
 DB_FILE_PATH = f"{HERE}/jupyter_scheduler/tests/testdb.sqlite"
 DB_URL = f"sqlite:///{DB_FILE_PATH}"
 
+TEST_ROOT_DIR = f"{HERE}/jupyter_scheduler/tests/test_root_dir"
+
```
dlqqq, line 17: 1. `os.path.join()` should be used to construct file paths, as the directory delimiter can be platform-dependent.
    - Windows uses backslashes instead of forward slashes, e.g. `C:\some_dir\child_dir`.
2. When using directories in tests, we should not commit the test dirs into the source repo, as this doesn't scale well when more tests are added. Instead, you should use the [`tmp_path` fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) provided by `pytest`. When used, this will create a new temporary directory (under `/tmp` on Linux) and return the path to that temporary directory. 

    So here you would want to replace `TEST_ROOT_DIR` with something like:

    ```python
    @pytest.fixture
    def scheduler_root_dir(tmp_path):
        return os.path.join(tmp_path, 'workspace_root')
    ```

For `DB_FILE_PATH`, you should use the `jp_data_dir` fixture provided by `jupyter_server.pytest_plugin`. This fixture is already under `tmp_path`.

jupyter_scheduler/tests/test_execution_manager.py
```diff
@@ -1,50 +1,55 @@
...
 from jupyter_scheduler.executors import DefaultExecutionManager
-from jupyter_scheduler.orm import Base, Job
+from jupyter_scheduler.orm import Job
 
-NOTEBOOK_DIR = Path(__file__).resolve().parent / "test_staging_dir" / "job-3"
+JOB_ID = "69856f4e-ce94-45fd-8f60-3a587457fce7"
 NOTEBOOK_NAME = "side_effects.ipynb"
+SIDE_EFECT_FILE_NAME = "output_side_effect.txt"
+
+NOTEBOOK_DIR = Path(__file__).resolve().parent / "test_staging_dir" / "job-4"
```
dlqqq, line 15: We should have a fixture for the staging directory which uses the default staging directory name in Jupyter Scheduler:

```py
@pytest.fixture
def staging_dir(jp_data_dir):
    return os.path.join(jp_data_dir, "scheduler_staging_area")
```

For the side effect test:

1. Move `output_side_effect.txt` and `side_effects.ipynb` under a new directory containing static files used in tests, e.g. `jupyter_scheduler/tests/static`.
2. Write a new `staging_dir_with_side_effects` fixture that copies these files to `staging_dir`, and returns a two-tuple of the input URI and the side effect file. So this fixture should return `('side_effects.ipynb', 'output_side_effect.txt')`

jupyter_scheduler/tests/test_execution_manager.py
```diff
@@ -1,50 +1,55 @@
...
-    db_session = sessionmaker(bind=engine)
-    with db_session() as session:
+@pytest.fixture
+def load_job(jp_scheduler_db):
+    with jp_scheduler_db() as session:
         job = Job(
             runtime_environment_name="abc",
             input_filename=NOTEBOOK_NAME,
-            job_id="123",
+            job_id=JOB_ID,
```
dlqqq, line 26: Instead of hard-coding the job ID, you should modify this fixture to return the ID of the job record it created. This is more similar to how jobs are created by the Scheduler backend.

jupyter_scheduler/tests/test_scheduler.py
```diff
@@ -39,6 +40,49 @@ def test_create_job_definition(jp_scheduler):
         assert "hello world" == definition.name
 
 
+def test_create_job_definition_with_input_folder(jp_scheduler):
+    job_definition_id = jp_scheduler.create_job_definition(
+        CreateJobDefinition(
+            input_uri="job-5/import-helloworld.ipynb",
```
dlqqq, line 46: Similar to my previous comment, you can replace this with a `input_file_with_input_dir` fixture that moves `import-helloworld.ipynb` and `a/b/helloworld.txt` to `scheduler_root_dir`, and then return the tuple `('import-helloworld.ipynb', 'a/b/helloworld.txt')`.

dlqqq (changes requested):
Thank you for contributing this! I left some feedback below. The feedback mostly centers around using fixtures instead of hard-coded paths, so these fixtures can be easily re-used in future tests.

dlqqq (approved):
In a 1:1 conversation, @andrii-i raised a good point that my feedback falls outside the scope of this PR, and that it would more appropriate to address this feedback for all tests (not just these 2 test files) in a separate PR. I'm tracking my previous feedback in a new issue: #514.

Awesome work!

andrii-i:
Thank you for looking into this @dlqqq

PR #516, opened 2024-05-02
Use pytest temporary folders and fixtures to create test file hierarchy at test time

### LINKED ISSUE #514: Prefer fixtures over hard-coded paths in tests
Thank you for contributing this! I left some feedback below. The feedback mostly centers around using fixtures instead of hard-coded paths, so these fixtures can be easily re-used in future tests.

_Originally posted by @dlqqq in https://github.com/jupyter-server/jupyter-scheduler/pull/513#pullrequestreview-2032030320_

andrii-i:
Improves Jupyter Scheduler pytest testing suite:
* Provide root and staging folders as fixtures using temporary folders instead of hard-coding paths in tests and storing folders in the codebase
* Store all test files in `static` folder, create file hierarchy to be tested at test time using temp folders and fixtures
* Make test db fixture return a session so that there is no need to instantiate a session every time

Fixes #514.


JasonWeill (approved):
I took a look at this change and I don't have any suggestions for you. Thanks!

conftest.py
```diff
@@ -1,27 +1,71 @@
...
 
-TEST_ROOT_DIR = f"{HERE}/jupyter_scheduler/tests/test_root_dir"
+@pytest.fixture(scope="session")
+def static_test_files_dir():
+    return Path(__file__).parent.resolve() / "jupyter_scheduler" / "tests" / "static"
 
 
 @pytest.fixture
-def jp_server_config(jp_server_config):
+def jp_scheduler_root_dir(tmp_path):
```
dlqqq, line 20: Can you add typing annotations to the return types of these fixtures? Some of these return `pathlib.Path` objects, and others return strings. Even though IDE intellisense doesn't really work for Pytest fixtures (at least on my VSCode instance), I think indicating the return type would be very helpful to future devs adding tests.

conftest.py
```diff
@@ -1,27 +1,71 @@
...
+    db_file_path = jp_scheduler_staging_dir / "scheduler.sqlite"
+    return f"sqlite:///{db_file_path}"
+
+
+@pytest.fixture
+def jp_scheduler_db(jp_scheduler_db_url):
+    engine = create_engine(jp_scheduler_db_url, echo=False)
+    Base.metadata.create_all(engine)
+    Session = sessionmaker(bind=engine)
+    return Session()
```
dlqqq, line 51: The DB connection should be closed after every test. You can define cleanup code by using the `yield` keyword instead of `return` here, e.g.

```py
...
session = Session()
yield session
session.close()
```

See: https://docs.pytest.org/en/latest/how-to/fixtures.html#yield-fixtures-recommended

dlqqq (changes requested):
Thanks @andrii-i! Left a few points of feedback below.

conftest.py
andrii-i: Agree. Added type annotations for all fixtures that return Paths, strings, or their tuples. I've also added assertions of `tmp_path` and similar as `Path` type to fixtures. At least on my instance, this makes IDE intellisense properly infer the return types increasing the usability for future devs even more.

dlqqq (approved):
Some of the fixtures are still missing types, but more importantly the path fixtures are now typed. The remaining type hints can be addressed in a separate PR.

Awesome work @andrii-i! Approved.

PR #520, opened 2024-05-20
Add database schema update and database migration logic

### LINKED ISSUE #519: Migration scripts? `no such column: jobs.package_input_folder`
v2.6.0 release some time ago added a new column `package_input_folder` https://github.com/jupyter-server/jupyter-scheduler/pull/510 but after upgrade it seems that the extension fails for me with:

![image](https://github.com/jupyter-server/jupyter-scheduler/assets/5832902/3d141a9c-c30c-4b96-a849-6b5ea65ffe6f)

## Description

```
│         return self._iter().one()                                                                                                                                                                                                                                                                                                                                                                                                      │
│                ^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                            │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/orm/query.py", line 2916, in _iter                                                                                                                                                                                                                                                                                                                         │
│         result = self.session.execute(                                                                                                                                                                                                                                                                                                                                                                                                 │
│                  ^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                 │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/orm/session.py", line 1717, in execute                                                                                                                                                                                                                                                                                                                     │
│         result = conn._execute_20(statement, params or {}, execution_options)                                                                                                                                                                                                                                                                                                                                                          │
│                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                          │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1710, in _execute_20                                                                                                                                                                                                                                                                                                                 │
│         return meth(self, args_10style, kwargs_10style, execution_options)                                                                                                                                                                                                                                                                                                                                                             │
│                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                             │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/sql/elements.py", line 334, in _execute_on_connection                                                                                                                                                                                                                                                                                                      │
│         return connection._execute_clauseelement(                                                                                                                                                                                                                                                                                                                                                                                      │
│                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                      │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1577, in _execute_clauseelement                                                                                                                                                                                                                                                                                                      │
│         ret = self._execute_context(                                                                                                                                                                                                                                                                                                                                                                                                   │
│               ^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                                                                   │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1953, in _execute_context                                                                                                                                                                                                                                                                                                            │
│         self._handle_dbapi_exception(                                                                                                                                                                                                                                                                                                                                                                                                  │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 2134, in _handle_dbapi_exception                                                                                                                                                                                                                                                                                                     │
│         util.raise_(                                                                                                                                                                                                                                                                                                                                                                                                                   │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/util/compat.py", line 211, in raise_                                                                                                                                                                                                                                                                                                                       │
│         raise exception                                                                                                                                                                                                                                                                                                                                                                                                                │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1910, in _execute_context                                                                                                                                                                                                                                                                                                            │
│         self.dialect.do_execute(                                                                                                                                                                                                                                                                                                                                                                                                       │
│       File "/opt/conda/envs/default/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute                                                                                                                                                                                                                                                                                                                │
│         cursor.execute(statement, parameters)                                                                                                                                                                                                                                                                                                                                                                                          │
│     sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: jobs.package_input_folder                                                                                                                                                                                                                                                                                                                              │
│     [SQL: SELECT count(*) AS count_1                                                                                                                                                                                                                                                                                                                                                                                                   │
│     FROM (SELECT jobs.runtime_environment_name AS jobs_runtime_environment_name, jobs.runtime_environment_parameters AS jobs_runtime_environment_parameters, jobs.compute_type AS jobs_compute_type, jobs.input_filename AS jobs_input_filename, jobs.output_formats AS jobs_output_formats, jobs.name AS jobs_name, jobs.tags AS jobs_tags, jobs.parameters AS jobs_parameters, jobs.email_notifications AS jobs_email_notifications, │
│     FROM jobs) AS anon_1]                                                                                                                                                                                                                                                                                                                                                                                                              │
│     (Background on this error at: https://sqlalche.me/e/14/e3q8)                                                                                                                                                                                                                                                                                                                                                                       │

```

## Reproduce

1. run some jobs in previous versions
2. upgrade
3, try to open the extension after upgrade

## Expected behavior

Either:
- there is a migration script
- there is an autofixer for database schemas

## Context

- Jupyter Server version: 4.x

andrii-i:
Add database schema update and database migration logic, test cases. 

With this PR, if database schema is extended between versions, schema is automatically updated and relevant tables are altered accordingly preventing errors. 

Fixes #519.

### DIFF
--- a/jupyter_scheduler/orm.py
+++ b/jupyter_scheduler/orm.py
@@ -114,24 +114,28 @@ class JobDefinition(CommonColumns, Base):
     active = Column(Boolean, default=True)
 
 
-def create_tables(db_url, drop_tables=False, Base=Base):
-    engine = create_engine(db_url)
+def update_db_schema(engine, Base):
     inspector = inspect(engine)
 
     with engine.connect() as connection:
         for table_name, model in Base.metadata.tables.items():
             if inspector.has_table(table_name):
-                columns_in_db = inspector.get_columns(table_name)
-                columns_in_db_names = {col["name"] for col in columns_in_db}
+                columns_db = inspector.get_columns(table_name)
+                columns_db_names = {col["name"] for col in columns_db}
 
-                for column_name, column in model.c.items():
-                    if column_name not in columns_in_db_names:
-                        column_type = str(column.type.compile(dialect=engine.dialect))
-                        nullable = "NULL" if column.nullable else "NOT NULL"
-                        alter_stmt = text(
-                            f"ALTER TABLE {table_name} ADD COLUMN {column_name} {column_type} {nullable}"
+                for column_model_name, column_model in model.c.items():
+                    if column_model_name not in columns_db_names:
+                        column_type = str(column_model.type.compile(dialect=engine.dialect))
+                        nullable = "NULL" if column_model.nullable else "NOT NULL"
+                        alter_statement = text(
+                            f"ALTER TABLE {table_name} ADD COLUMN {column_model_name} {column_type} {nullable}"
                         )
-                        connection.execute(alter_stmt)
+                        connection.execute(alter_statement)
+
+
+def create_tables(db_url, drop_tables=False, Base=Base):
+    engine = create_engine(db_url)
+    update_db_schema(engine, Base)
 
     try:
         if drop_tables:
--- a/jupyter_scheduler/orm.py
+++ b/jupyter_scheduler/orm.py
@@ -122,22 +122,28 @@ class JobDefinition(CommonColumns, Base):
 
 def update_db_schema(engine, Base):
     inspector = inspect(engine)
+    alter_statements = []
 
+    for table_name, model in Base.metadata.tables.items():
+        if not inspector.has_table(table_name):
+            continue
+        columns_db = inspector.get_columns(table_name)
+        columns_db_names = {col["name"] for col in columns_db}
+
+        for column_model_name, column_model in model.c.items():
+            if column_model_name in columns_db_names:
+                continue
+            column_type = str(column_model.type.compile(dialect=engine.dialect))
+            alter_statement = text(
+                f"ALTER TABLE {table_name} ADD COLUMN {column_model_name} {column_type} NULL"
+            )
+            alter_statements.append(alter_statement)
+
+    if not alter_statements:
+        return
     with engine.connect() as connection:
-        for table_name, model in Base.metadata.tables.items():
-            if inspector.has_table(table_name):
-                columns_db = inspector.get_columns(table_name)
-                columns_db_names = {col["name"] for col in columns_db}
-
-                for column_model_name, column_model in model.c.items():
-                    if column_model_name in columns_db_names:
-                        continue
-                    if column_model_name not in columns_db_names:
-                        column_type = str(column_model.type.compile(dialect=engine.dialect))
-                        alter_statement = text(
-                            f"ALTER TABLE {table_name} ADD COLUMN {column_model_name} {column_type} NULL"
-                        )
-                        connection.execute(alter_statement)
+        for alter_statement in alter_statements:
+            connection.execute(alter_statement)
 
 
 def create_tables(db_url, drop_tables=False, Base=Base):
--- a/jupyter_scheduler/orm.py
+++ b/jupyter_scheduler/orm.py
@@ -4,8 +4,9 @@ from sqlite3 import OperationalError
 from uuid import uuid4
 
 import sqlalchemy.types as types
-from sqlalchemy import Boolean, Column, Integer, String, create_engine
+from sqlalchemy import Boolean, Column, Integer, String, create_engine, inspect
 from sqlalchemy.orm import declarative_base, declarative_mixin, registry, sessionmaker
+from sqlalchemy.sql import text
 
 from jupyter_scheduler.models import EmailNotifications, Status
 from jupyter_scheduler.utils import get_utc_timestamp
@@ -112,8 +113,25 @@ class JobDefinition(CommonColumns, Base):
     active = Column(Boolean, default=True)
 
 
-def create_tables(db_url, drop_tables=False):
+def create_tables(db_url, drop_tables=False, Base=Base):
     engine = create_engine(db_url)
+    inspector = inspect(engine)
+
+    with engine.connect() as connection:
+        for table_name, model in Base.metadata.tables.items():
+            if inspector.has_table(table_name):
+                columns_in_db = inspector.get_columns(table_name)
+                columns_in_db_names = {col["name"] for col in columns_in_db}
+
+                for column_name, column in model.c.items():
+                    if column_name not in columns_in_db_names:
+                        column_type = str(column.type.compile(dialect=engine.dialect))
+                        nullable = "NULL" if column.nullable else "NOT NULL"
+                        alter_stmt = text(
+                            f"ALTER TABLE {table_name} ADD COLUMN {column_name} {column_type} {nullable}"
+                        )
+                        connection.execute(alter_stmt)
+
     try:
         if drop_tables:
             Base.metadata.drop_all(engine)
--- a/jupyter_scheduler/orm.py
+++ b/jupyter_scheduler/orm.py
@@ -1,5 +1,4 @@
 import json
-import os
 from sqlite3 import OperationalError
 from uuid import uuid4
 
@@ -92,6 +91,7 @@ class CommonColumns:
 
 class Job(CommonColumns, Base):
     __tablename__ = "jobs"
+    __table_args__ = {"extend_existing": True}
     job_id = Column(String(36), primary_key=True, default=generate_uuid)
     job_definition_id = Column(String(36))
     status = Column(String(64), default=Status.STOPPED)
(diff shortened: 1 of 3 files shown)


JasonWeill (approved):

jupyter_scheduler/orm.py
```diff
@@ -104,6 +105,7 @@ class Job(CommonColumns, Base):
 
 class JobDefinition(CommonColumns, Base):
     __tablename__ = "job_definitions"
+    __table_args__ = {"extend_existing": True}
```
dlqqq, line 108: I had to do some testing to understand why this is necessary if we are defining `update_db_schema()`. Essentially, this argument is needed to extend the table _metadata_ (e.g. column names & types) associated with the table. If this line is omitted in `test_orm.py`, the test fails with the error:

```
sqlalchemy.exc.InvalidRequestError: Table 'jobs' is already defined for this MetaData instance.  Specify 'extend_existing=True' to redefine options and columns on an existing Table object.
```

So both the table migration (defined in `update_db_schema()`) and table _metadata_ migration (defined here by setting `__table_args__ = {"extend_existing": True}`) must be performed when a new column is added. 😵 

No action needed. I just felt it was necessary to call this out to other readers in this review.

jupyter_scheduler/orm.py
```diff
@@ -112,8 +114,29 @@ class JobDefinition(CommonColumns, Base):
...
+                columns_db = inspector.get_columns(table_name)
+                columns_db_names = {col["name"] for col in columns_db}
+
+                for column_model_name, column_model in model.c.items():
+                    if column_model_name not in columns_db_names:
+                        column_type = str(column_model.type.compile(dialect=engine.dialect))
+                        nullable = "NULL" if column_model.nullable else "NOT NULL"
+                        alter_statement = text(
+                            f"ALTER TABLE {table_name} ADD COLUMN {column_model_name} {column_type} {nullable}"
+                        )
```
dlqqq, line 132: Two comments:

1. We shouldn't be using f-strings to pass in arguments to a SQL statement, as this doesn't sanitize the arguments beforehand. While the risk of a SQL injection is very low here, we should still sanitize all our SQL statements to negate the risk entirely. Here is the recommended syntax in SQLAlchemy: https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.text

2. This statement a) always fails for `NOT NULL` columns, and b) never respects the default. The reason for this is that a) `NOT NULL` columns require a default value, and b) the default needs to be explicitly specified via a `DEFAULT` clause, e.g. `ALTER TABLE PERSON ADD IS_ACTIVE VARCHAR2(1) DEFAULT 'N'`. Since this SQL statement lacks a `DEFAULT` clause, the default value specified in the column is never passed to the `ALTER` statement, meaning that non-nullable columns always fail to be created.

    The reason this went unnoticed is because `nullable` defaults to `True` in SQLAlchemy, and this argument is not specified in the `package_input_folder` and `packaged_input_files` columns added to the table.

    I think it's fine if we avoid the complexity of supporting `NOT NULL` columns, i.e. require all future columns we add to the tables to be nullable. Therefore, to address this concern, I recommend:
    
    - Explicitly document the fact that all future columns must be nullable and that default values are ignored. This should be placed in a comment at the bottom of the `CommonColumns`, `Job`, and `JobDefinition` table definitions above `update_db_schema()`.
    - Always use `NULL` in the highlighted SQL statement above.

jupyter_scheduler/orm.py
```diff
@@ -112,8 +114,29 @@ class JobDefinition(CommonColumns, Base):
...
+    inspector = inspect(engine)
+
+    with engine.connect() as connection:
+        for table_name, model in Base.metadata.tables.items():
+            if inspector.has_table(table_name):
+                columns_db = inspector.get_columns(table_name)
+                columns_db_names = {col["name"] for col in columns_db}
+
+                for column_model_name, column_model in model.c.items():
+                    if column_model_name not in columns_db_names:
```
dlqqq, line 127: We can remove one level of indentation by replacing this with:

```py
if column_model_name in columns_db_names:
    continue
...
```

dlqqq (dismissed):
@andrii-i Thank you for working on this! Left some comments below.

jupyter_scheduler/orm.py
andrii-i: Implemented recommendation from the point 2 to ignore null-ability and default values during the db migration, document the fact via comments in orm models.

Re point 1:
Documentation on this is sparse. In SQLite parameter “specifies a placeholder in the expression for a value that is filled in at runtime” (source: https://www.sqlite.org/lang_expr.html#varparam). Table names, table column names, and column types are database identifiers (https://www.sqlite.org/quirks.html#keywords_can_often_be_used_as_identifiers) which is a different “type” from values. Identifiers are used/interpreted differently by SQLite engine.
Here are a couple of good stack overflow threads on the topic:
https://stackoverflow.com/questions/3247183/variable-table-name-in-sqlite
https://stackoverflow.com/questions/1274432/sqlite-parameters-not-allowing-tablename-as-parameter
So when I try to parametrize database identifiers like in code below, syntax error is thrown.
Code:
```
                    alter_statement = text(
                        "ALTER TABLE :table_name ADD COLUMN :column_model_name :column_type NULL"
                    ).bindparams(
                        table_name=table_name,
                        column_model_name=column_model_name,
                        column_type=column_type,
                    )
                    connection.execute(alter_statement)
```
Error:
```
self = <sqlalchemy.dialects.sqlite.pysqlite.SQLiteDialect_pysqlite object at 0x113638c50>, cursor = <sqlite3.Cursor object at 0x113640640>, statement = 'ALTER TABLE ? ADD COLUMN ? ? NULL'
parameters = ('jobs', 'new_column', 'INTEGER'), context = <sqlalchemy.dialects.sqlite.base.SQLiteExecutionContext object at 0x113639d50>

    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "?": syntax error
E       [SQL: ALTER TABLE ? ADD COLUMN ? ? NULL]
E       [parameters: ('jobs', 'new_column', 'INTEGER')]
E       (Background on this error at: https://sqlalche.me/e/20/e3q8)

../../miniconda3/envs/j-sch-t/lib/python3.11/site-packages/sqlalchemy/engine/default.py:924: OperationalError
================================================================================= short test summary info =================================================================================
FAILED jupyter_scheduler/tests/test_orm.py::test_create_tables_with_new_column - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "?": syntax error
[SQL: ALTER TABLE ? ADD COLUMN ? ? NULL]
[parameters: ('jobs', 'new_column', 'INTEGER')]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
```

Therefore ignoring the callout 1 as discussed during a chat with @dlqqq on the topic.

JasonWeill (approved):

jupyter_scheduler/orm.py
```diff
@@ -100,20 +103,47 @@ class Job(CommonColumns, Base):
...
     active = Column(Boolean, default=True)
+    # All new columns added to this table must be nullable to ensure compatibility during database migrations.
+    # Any default values specified for new columns will be ignored during the migration process.
 
 
-def create_tables(db_url, drop_tables=False):
+def update_db_schema(engine, Base):
+    inspector = inspect(engine)
+
+    with engine.connect() as connection:
```
3coins, line 126: nit: Rather than wrapping the logic with the connection, it is cleaner to collect all the alter statements and execute this separately after this in a separate block.
```python
# Rest of the logic here

with engine.connect() as connection:
    for alter_statement in alter_statements:
        connection.execute(alter_statement)
```

andrii-i: Good point, it is cleaner this way. I also added a check to ensure that `alter_statements` is not empty before instantiating the connection so this would also optimize a number of db connections.

3coins (approved):
Looks good! 🚀

andrii-i:
Thank you everyone for reviews.

PR #521, opened 2024-05-21
Update to SQLAlchemy 2.x

### LINKED ISSUE #478: Upgrade to SQLAlchemy 2.x
Hi all,

Pandas is now depending on SQLAlchemy 2.x
When using pandas and jupyterlab at the same time, which is a very common situation, it may case some conflicts.

https://github.com/jupyter-server/jupyter-scheduler/blob/1b50f81f966c9e0146f91b1606fd78519857527e/pyproject.toml#L36

andrii-i:
Jupyter Scheduler uses subset of SQLAlchemy functionality supported in both SQLAlchemy 1.x and 2.x.

Updating to SQLAlchemy 2.x would allow people who want to use Jupyter Scheduler and other packages that depend on SQLAlchemy 2.x in the same environment (for example, [pandas](https://github.com/pandas-dev/pandas/blob/986074b4901511a4e87a99854fc62f8c24bdbb71/requirements-dev.txt#L45)) to do so.

For reference: [SQLAlchemy 2.0 - Major Migration Guide ](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#the-1-4-2-0-migration-path).

Fixes #478.
Will also fix a SQLAlchemy 1.x depreciation warning failing CI in #520: https://github.com/jupyter-server/jupyter-scheduler/actions/runs/9179340732/job/25241262049?pr=520#step:6:463

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ details.
 > Starting with v2.7.0, Jupyter Scheduler requires SQLAlchemy 2.x instead of SQLAlchemy 1.x.
 
 > [!IMPORTANT]
-> JupyterLab 3 reached its end of maintenance date on May 15, 2024, anywhere on Earth. As a result, we will not backport new features to the v1 branch supporting JupyterLab 3 after this date. Fixes for critical issues will still be backported until December 31, 2024. If you are still using JupyterLab 3, we strongly encourage you to **upgrade to JupyterLab 4 as soon as possible**. For more information, see [JupyterLab 3 end of maintenance](https://blog.jupyter.org/jupyterlab-3-end-of-maintenance-879778927db2) on the Jupyter Blog.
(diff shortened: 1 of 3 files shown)


srdas (commented):
Reviewed the change and also tested it. Works as expected.  
image

andrii-i:
Thank you @JasonWeill, @srdas.

README.md
```diff
@@ -17,6 +17,9 @@ details.
 - JupyterLab 4.x (for newer Jupyter Scheduler versions)
 - JupyterLab 3.x (for Jupyter Scheduler 1.x)
 
+> [!IMPORTANT]
+> Starting with v2.7.0, Jupyter Scheduler requires SQLAlchemy 2.x instead of SQLAlchemy 1.x.
+
```
JasonWeill, line 22: I know this line isn't modified as part of this PR, but line 24 can now start, "JupyterLab 3 reached its end of maintenance …".

JasonWeill (approved):
I'm a little concerned that we have two alert boxes right next to each other, but if we eventually phase out both of them, this won't be a problem in the long run.

README.md
andrii-i: Changed "will reach" to "reached" in both README and readthedocs.

PR #523, opened 2024-05-30
Emit telemetry event on "Run job with input folder" checkbox click

andrii-i:
Emit `create-job.options.package_input_folder.{check/uncheck}` on "Run job with input folder" checkbox click. Metric name is proposed because it is semantically in line with metrics already being emitted: [1](https://github.com/jupyter-server/jupyter-scheduler/blob/dbeb34fd357cc5f9bfad28cb1233e7041d25e24c/src/mainviews/create-job.tsx#L341), [2](https://github.com/jupyter-server/jupyter-scheduler/blob/dbeb34fd357cc5f9bfad28cb1233e7041d25e24c/src/mainviews/create-job.tsx#L598), [3](https://github.com/jupyter-server/jupyter-scheduler/blob/dbeb34fd357cc5f9bfad28cb1233e7041d25e24c/src/mainviews/create-job.tsx#L548-L552).

Note that by default [LogContext is empty](https://github.com/jupyter-server/jupyter-scheduler/blob/dbeb34fd357cc5f9bfad28cb1233e7041d25e24c/src/context.ts#L5-L9) so telemetry goes nowhere (Jupyter Scheduler operator needs to overwrite it with a custom logging function to actually use telemetry). So this PR is in line what we are already doing and does not expose Jupyter Scheduler users to anything.

### DIFF
--- a/src/components/input-folder-checkbox.tsx
+++ b/src/components/input-folder-checkbox.tsx
@@ -1,5 +1,5 @@
 import React, { ChangeEvent } from 'react';
-import { useTranslator } from '../hooks';
+import { useEventLogger, useTranslator } from '../hooks';
 
 import {
   Checkbox,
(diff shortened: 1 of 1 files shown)


sravyasdh:
LGTM! Thank you for adding this!

srdas (approved):
All looks good. Tested the branch as well and everything is working as expected.

PR #530, opened 2024-06-21
Use Dask instead of multiprocessing module

andrii-i:
Use Dask instead of multiprocessing module


andrii-i:
Kicking CI

jupyter_scheduler/extension.py
```diff
@@ -91,3 +92,25 @@ def initialize_settings(self):
...
+        """
+        Public method called by Jupyter Server when the server is stopping.
+        This calls the cleanup code defined in `self._stop_exception()` inside
+        an exception handler, as the server halts if this method raises an
+        exception.
+        """
+        try:
+            await self._stop_extension()
+        except Exception as e:
+            self.log.error("Jupyter Scheduler raised an exception while stopping:")
```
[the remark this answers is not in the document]
andrii-i: Thank you for looking into this. Error message in question indicates that Jupyter Scheduler stop_extension logic threw an error. Note that the next line `self.log.exception(e)` would log the actual error. From my point of view this provides an explanation with a sufficient level of details. 

Could you please elaborate on your suggestion? What kind of detail is missing? What kind of message would you suggest?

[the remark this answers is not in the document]
srdas: Agree, that is all that is needed. Thanks for clarifying.

conftest.py
```diff
@@ -1,4 +1,5 @@
 from pathlib import Path
+from unittest.mock import AsyncMock
```
JasonWeill, line 2: This is the only change to this file, and `AsyncMock` isn't used anywhere else in this PR. Should it remain?

jupyter_scheduler/scheduler.py
```diff
@@ -381,6 +382,12 @@ def get_local_output_path(
         else:
             return os.path.join(self.root_dir, self.output_directory, output_dir_name)
 
+    async def stop_extension(self):
+        """
+        Placeholder method for a cleanup code to run when the server is stopping.
+        """
+        pass
```
JasonWeill, line 389: Do we need to keep this method if it does nothing?

JasonWeill (approved):
Generally looking good, just a couple of small non-blocking comments.

jupyter_scheduler/scheduler.py
andrii-i: Yes. I [implement](https://github.com/andrii-i/jupyter-scheduler/blob/4c102d14415aeb6f29b0f1022fddc83c4b0058d6/jupyter_scheduler/scheduler.py#L788-L794) this method in `Scheduler.stop_extension`. Note that `BaseScheduler` is a[ base class / klass for schedulers](https://github.com/andrii-i/jupyter-scheduler/blob/4c102d14415aeb6f29b0f1022fddc83c4b0058d6/jupyter_scheduler/extension.py#L56-L61), essentially an interface.

To correctly terminate Dask and avoid JupyterLab process not shutting down immediately on interrupt as we've seen in Jupyter AI, I need to [call](https://github.com/andrii-i/jupyter-scheduler/blob/4c102d14415aeb6f29b0f1022fddc83c4b0058d6/jupyter_scheduler/extension.py#L96-L116) `scheduler.stop_extension` from `SchedulerApp.stop_extension` (a public method inherited from `ExtensionApp` called by Jupyter Server when the server is stopping). Note that `scheduler` in this case is some subclass of `BaseScheduler`.

So `BaseScheduler` base class / klass / interface needs to have a placeholder `stop_extension` method that does nothing by default and that subclasses can implement `stop_extension` logic in.

andrii-i: Yes. `BaseScheduler` is a [base class / klass for schedulers](https://github.com/andrii-i/jupyter-scheduler/blob/4c102d14415aeb6f29b0f1022fddc83c4b0058d6/jupyter_scheduler/extension.py#L56-L61), essentially an interface for all schedulers [instantiated and used](https://github.com/andrii-i/jupyter-scheduler/blob/4c102d14415aeb6f29b0f1022fddc83c4b0058d6/jupyter_scheduler/extension.py#L77) by `SchedulerApp`. `BaseScheduler.stop_extension` defines scheduler cleanup / stop_extension logic (if any) and is [called](https://github.com/andrii-i/jupyter-scheduler/blob/4c102d14415aeb6f29b0f1022fddc83c4b0058d6/jupyter_scheduler/extension.py#L96-L116) from `SchedulerApp.stop_extension`. 
`SchedulerApp.stop_extension` is a public method inherited from `ExtensionApp` called by Jupyter Server when the server is stopping.  

`stop_extension` is [implemented](https://github.com/andrii-i/jupyter-scheduler/blob/4c102d14415aeb6f29b0f1022fddc83c4b0058d6/jupyter_scheduler/scheduler.py#L788-L794)  in default `Scheduler.stop_extension` to correctly terminate Dask and avoid JupyterLab process not shutting down immediately on interrupt as for example we've seen in Jupyter AI.

andrii-i:
Superseded by #553

PR #535, opened 2024-07-09
Updated integration tests workflow

krassowski:
Updated integration tests workflow

### DIFF
--- a/.github/workflows/update-integration-tests.yml
+++ b/.github/workflows/update-integration-tests.yml
@@ -1,11 +1,6 @@
 name: Update Playwright Snapshots
 
 on:
-  workflow_dispatch:
-    inputs:
-      number:
-        description: 'PR number'
-        required: true
   issue_comment:
     types: [created, edited]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/update-integration-tests.yml
```diff
@@ -1,11 +1,6 @@
 name: Update Playwright Snapshots
 
 on:
-  workflow_dispatch:
-    inputs:
-      number:
-        description: 'PR number'
-        required: true
```
krassowski, line 8: You may want to keep this, but the body of action would probably need some more changes. I recommend merging as-is and restoring the `workflow_dispatch` later.

andrii-i (approved):
This updates "Update Playwright Snapshots" workflow in line with JupyterLab https://github.com/jupyterlab/jupyterlab/blob/main/.github/workflows/galata-update.yml. LGTM. Thank you for creating this PR @krassowski.

krassowski:
Thanks @andrii-i! I do not have write access on jupyter-server so I have leave it to you or someone else from your team to merge it :)

Issue #558, opened 2024-11-13
Jupyter Scheduler 2.10.0 Source Distribution is not available


andrii-i:
## Description

Jupyter Scheduler 2.10.0 Source Distribution is not available as its [upload failed](https://github.com/jupyter-server/jupyter-scheduler/actions/runs/11808986203/job/32898512565#step:5:372) due to PyPI source distributions size limits (~150 Mb) due to dependencies being pulled into the `tar` file (should not happen). 

Built distribution upload went through and is available at https://pypi.org/project/jupyter-scheduler/2.10.0/#files and for installation via `pip` for systems and environments that require source distribution.

Impact: systems and environments that require source distribution.

## How to reproduce

CI failure: https://github.com/jupyter-server/jupyter-scheduler/actions/runs/11808986203/job/32898512565#step:5:372
PyPI packages available: https://pypi.org/project/jupyter-scheduler/2.10.0/#files

## Expected behavior

Source distribution of the normal size is available in PyPI.

dlqqq:
For reference, here is the relevant log excerpt:

```
WARNING  Error during upload. Retry with the --verbose option for more details.
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
         File too large. Limit for project 'jupyter-scheduler' is 100 MB. See
         https://pypi.org/help/#file-size-limit for more information.
```

andrii-i:
Jupyter Scheduler 2.10.0 npm package is now available at https://www.npmjs.com/package/@jupyterlab/scheduler, Source Distribution is now available at PyPI https://pypi.org/project/jupyter-scheduler/2.10.0/#files. 

Let's use this issue to track the need to understand why Jupyter Scheduler 2.10.0 Source Distribution tar was built too large causing PyPI upload failure and to prevent it happening in the next release. 

jupyter_releaser issue on the topic: https://github.com/jupyter-server/jupyter_releaser/issues/592

krassowski:
> Try building jupyter scheduler PyPI source distribution locally with `jupyter-releaser build-python`, see its size (>100 Mb)

Out of curiosity, do you know why it produces so big a distribution? From a quick look it seems that you might be missing:

```toml
[tool.jupyter-releaser.hooks]
before-build-python = ["jlpm clean:all"]
```

in the `pyproject.toml` but that's just a guess.

andrii-i:
[...]

PR #561, opened 2024-11-15
Debug jupyter-releaser issue

krassowski:
## References

https://github.com/jupyter-server/jupyter-scheduler/issues/558

## Code changes

Just blindly checking if changing this could have an impact

## User-facing changes

None

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,17 +128,5 @@
     "extension": true,
     "outputDir": "jupyter_scheduler/labextension",
     "schemaDir": "schema"
-  },
-  "jupyter-releaser": {
-    "hooks": {
-      "before-build-npm": [
-        "python -m pip install jupyterlab~=4.0",
-        "jlpm",
-        "jlpm build:prod"
-      ],
-      "before-build-python": [
-        "jlpm clean:all"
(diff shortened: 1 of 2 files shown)


krassowski:
Oh, `check-release` is not running on PRs until they are closed due to this commit: https://github.com/jupyter-server/jupyter-scheduler/commit/51ca23c735cd340f47d1309ce9b2f7ded6883a80

krassowski:
This exposed an error `ValueError: @jupyterlab/scheduler is missing 'main' file lib/index.js`

krassowski:
Nope, this did not help at all:

image

krassowski:
It is quite intriguing as other repositories do not experience this issue, but I cannot put any more time into it this week.

I am able to reproduce it locally too. I enabled logging by editing jupyter-releaser code locally but nothing come up in there.

A laborious but possibly more fruitful approach to debugging it would be starting from a fresh cookiecutter and moving over parts of jupyter-scheduler to see what causes this behaviour.

andrii-i:
Thank you for looking into this @krassowski.

PR #587, opened 2025-07-23
Add cell execution tracking during notebook execution

### LINKED ISSUE #586: Track completed cell progress during notebook execution
### Problem

  When executing notebooks through the Jupyter Scheduler, users have no visibility into the real-time progress of long-running jobs. This creates frustration because:

  * Users cannot tell if a job is actively progressing or stuck on a particular cell
  * There's no way to estimate how much work remains for a running job
  * Users resort to checking logs or stopping jobs unnecessarily due to lack of progress feedback

  ### Proposed solution

I propose we add a `completed_cells` field to the Job model to track the number of cells executed. This will be updated after every cell execution by leveraging the `code_cells_executed` in the nbclient NotebookClient ([ref](https://github.com/jupyter/nbclient/blob/main/nbclient/client.py#L473)) and updating the database row for the job with it. We can do this by adapting the ExecutePreprocessor used today (which inherits from the NotebookClient):

```python
class TrackingExecutePreprocessor(ExecutePreprocessor):
    """Custom ExecutePreprocessor that tracks completed cells and updates the database"""
    
    def __init__(self, db_session, job_id, **kwargs):
        super().__init__(**kwargs)
        self.db_session = db_session
        self.job_id = job_id
    
    def preprocess_cell(self, cell, resources, index):
        """
        Override to track completed cells in the database.
        Calls the superclass implementation and then updates the database.
        """
        # Call the superclass implementation
        cell, resources = super().preprocess_cell(cell, resources, index)
        
        # Update the database with the current count of completed cells
            session.query(Job).filter(Job.job_id == self.job_id).update(
                {"completed_cells": self.code_cells_executed}
            )
            session.commit()
        
        return cell, resources
```

  In total, we make the following changes:
1. Model update to the Job model to add the `completed_cells` field
2. Implement the `TrackingExecutePreprocessor` as described above
3. Update the GET `jobs/{job_id}` to expose the completed cells in the response body
4. Update the PATCH `jobs/{job_id}` to allow manual patching of the completed cells value if needed

  This enables users to:
  - Monitor job progress in real-time through the API. We can build this into the front-end component as a separate task.
  - Make informed decisions about stopping or continuing long-running jobs

As a bonus, stopped jobs will leave the last completed cell in the field, allowing users to identify which cell a job failed on for faster debugging.

  ### Additional context

  This feature addresses a common need in notebook execution systems. Similar implementations exist in:

  * **Papermill**: Tracks notebook execution progress with cell-level granularity. Also includes a progress tracker in stdout (although we don't have to do this)
  * **Google Colab**: Shows real-time cell execution progress in the UI

agupta01:
## References

Resolves #586 

## Code changes

This PR adds a `completed_cells` field to the Job model to track the number of cells executed during notebook execution in real-time. The implementation includes:

  Model Changes:
  - Added nullable `completed_cells` column to the Job table in ORM (`jupyter_scheduler/orm.py`)
  - Added completed_cells field to DescribeJob and UpdateJob models (`jupyter_scheduler/models.py`)

  Execution Tracking:
  - Added `JobFeature.track_cell_execution` feature flag to enable/disable cell tracking
  - Modified `DefaultExecutionManager` to use nbconvert's native `on_cell_executed` hook. Note that this was found to be cleaner than the subclassing approach defined in #586.
  - Hook updates the database with ep.code_cells_executed after each cell execution

## User-facing changes

  - GET `/jobs/{job_id}` now returns the current count of completed cells
  - PATCH `/jobs/{job_id}` accepts completed_cells updates for manual corrections

None

### DIFF
--- a/jupyter_scheduler/executors.py
+++ b/jupyter_scheduler/executors.py
@@ -29,9 +29,7 @@ class ExecutionManager(ABC):
     _model = None
     _db_session = None
 
-    def __init__(
-        self, job_id: str, root_dir: str, db_url: str, staging_paths: Dict[str, str]
-    ):
+    def __init__(self, job_id: str, root_dir: str, db_url: str, staging_paths: Dict[str, str]):
         self.job_id = job_id
         self.staging_paths = staging_paths
         self.root_dir = root_dir
@@ -175,14 +173,10 @@ class DefaultExecutionManager(ExecutionManager):
         if new_files_set:
             with self.db_session() as session:
                 current_packaged_files_set = set(
-                    session.query(Job.packaged_files)
-                    .filter(Job.job_id == self.job_id)
-                    .scalar()
+                    session.query(Job.packaged_files).filter(Job.job_id == self.job_id).scalar()
                     or []
                 )
-                updated_packaged_files = list(
-                    current_packaged_files_set.union(new_files_set)
-                )
+                updated_packaged_files = list(current_packaged_files_set.union(new_files_set))
                 session.query(Job).filter(Job.job_id == self.job_id).update(
                     {"packaged_files": updated_packaged_files}
                 )
@@ -192,9 +186,7 @@ class DefaultExecutionManager(ExecutionManager):
         for output_format in job.output_formats:
             cls = nbconvert.get_exporter(output_format)
             output, _ = cls().from_notebook_node(notebook_node)
-            with fsspec.open(
-                self.staging_paths[output_format], "w", encoding="utf-8"
-            ) as f:
+            with fsspec.open(self.staging_paths[output_format], "w", encoding="utf-8") as f:
                 f.write(output)
 
     def supported_features(self) -> Dict[JobFeature, bool]:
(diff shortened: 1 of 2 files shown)


andrii-i (commented):
Great work @agupta01, thank you. This PR adds cell execution tracking feature in a high-quality and clean manner, db schema changes are done properly and therefore backwards-compatible.

Refactors of `validate()` and `supported_features()` to be classmethods makes sense to me and is a clear improvement / cleaner design since both methods don't need instance state.

Two concerns prevent immediate merge:
1. Enabling cell tracking by default makes every cell execution trigger a database write. This could impact production deployments, especially for users with large notebooks and limited compute and networking resources / configurations. Making this feature opt-in / disabled by default would remedy this.
2. Classmethod refactor breaks backwards compatibility for any custom `ExecutionManager` implementations and therefore is a breaking change. Refactoring these changes into a separate PR that would be merged after the final `2.x` release would remedy this.

Great work. Happy to merge once these changes are made. The cell tracking feature will be a great addition to Jupyter Scheduler.

agupta01:
@andrii-i thanks for your feedback. I have made the requested changes:

1. Cell tracking is now off by default, and is opt-in by setting `track_cell_execution` to True
2. Removed the classmethod refactoring in `ExecutionManager`, will open another PR with those changes as recommended.

Issue #90, opened 2022-10-04
Implement timeouts, retries for jobs


3coins:
Implement 
- **timeout_seconds** timeout in seconds for the job to complete
- **retry_on_timeout** whether to retry if a timeout has occurred
- **max_retries** no of times to attempt retry
- **min_retry_interval_millis** delay in milliseconds between retry attempts

PR #596, opened 2025-12-10
Support multiple backends

### LINKED ISSUE #597: Support multiple backends
### Problem

Jupyter Scheduler has a modular architecture that allows users to implement and swap in various backend components, enabling deep customization of scheduler functionality. However, only one backend can be active forcing users to choose a single backend for all jobs before Jupyter Scheduler launch.

### Proposed solution

* Support running multiple backends simultaneously
* Allow users to choose which backend to use during job creation

andrii-i:
## Summary

* Fixes #597, part of #599

Update Jupyter Scheduler to support scheduling and running arbitrary file types beyond notebooks. Multiple backends can now run simultaneously, allowing different file types to be handled by their respective backends within the same UI.

Introduces the `BaseBackend` abstraction, allowing backend authors to define capabilities declaratively (supported file types, output formats, scheduler/executor classes) with automatic discovery via entry points. Custom backends before could have been defined as sets of classes; this PR formalizes the pattern into a `BaseBackend` class with declarative configuration and automatic discovery via entry points.

### New `BaseBackend` abstraction

```
from jupyter_scheduler.base_backend import BaseBackend

class MyBackend(BaseBackend):
    id = "my_backend"
    name = "My Custom Backend"
    description = "Execute notebooks on my infrastructure"
    scheduler_class = "my_package.scheduler:MyScheduler"
    execution_manager_class = "my_package.executors:MyExecutionManager"
    file_extensions = ["ipynb"]
    output_formats = [{"id": "ipynb", "label": "Notebook"}]
```

### Available backends discovery via new REST API endpoint
- New `GET /scheduler/backends` endpoint returns available backends with their capabilities

### Backends Provided by Default with Jupyter Scheduler
- `jupyter_server_nb` - Execute notebooks via nbconvert (refactoring of an existing notebook execution logic)
- `jupyter_server_py` - Execute Python scripts via local subprocess

### Backend Discovery via Entry Points
Backends are registered using Python entry points in pyproject.toml:

```
[project.entry-points."jupyter_scheduler.backends"]
my_backend = "my_package.backends:MyBackend"
```

### Backend Selection UI
- Backend picker dropdown in "Create Job" form (when multiple backends support the file type)
- Shows backend name and description
- Auto-selects based on file extension

### Configuration Options

Route legacy jobs (pre-3.0 UUID-only IDs) to a specific backend
```
c.SchedulerApp.legacy_job_backend = "jupyter_server_nb"
```

Set preferred backend per file extension
```
c.SchedulerApp.preferred_backends = {"ipynb": "k8s_backend"}
```

## Code Changes

New files:
- jupyter_scheduler/base_backend.py - Base class for backends
- jupyter_scheduler/backend_registry.py - Registry for managing backends
- jupyter_scheduler/backend_utils.py - backends discovery logic via entry points
- jupyter_scheduler/job_id.py - Job ID encoding (backend_id:uuid) logic
- src/util/backend-utils.ts - backend utilities

Modified:
- extension.py - backend discovery and initialization
- handlers.py - logic to route requests to correct backend
- create-job.tsx - added backend picker UI
- handler.ts - added new /scheduler/backends API endpoint

## User-facing changes

Screenshot 2025-12-09 at 9 34 09 PM

None. 

Pre-v3 pre-multiple-backend jobs have UUID-only IDs and are routed to `legacy_job_backend` which be default is set to pre-v3 local notebook execution logic. So without installing additional backends default installation works identically (single jupyter_server_nb backend).

## Testing

- All pre-existing tests pass 
- Added new tests including Playwright E2E tests for multi-backend UI with multiple mocked multiple backends


jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+            ext = input_uri.rsplit(".", 1)[-1].lower()
+
+        candidates = self._extension_map.get(ext, [])
+        if not candidates:
+            raise ValueError(f"No backend supports file extension '.{ext}'")
+
+        # 1. Explicit preference for this extension
+        preferred = self._preferred_backends.get(ext)
+        if preferred and preferred in candidates:
+            return self._backends[preferred]
```
dlqqq: `self._preferred_backends` is annotated with type `dict[str, str]` in the `__init__()` constructor. However, it seems like this expects the attribute to be of type `dict[str, BackendInstance]`. Which one is correct?

jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+                id=b.config.id,
+                name=b.config.name,
+                description=b.config.description,
+                file_extensions=b.config.file_extensions,
+                output_formats=b.config.output_formats,
+            )
+            for b in backends_sorted
+        ]
+
+    def list_backend_instances(self) -> List[BackendInstance]:
```
dlqqq: nit:

- rename `list_backends()` to `describe_backends()`
- rename `list_backend_instances()` to `list_backends()` or a property like `@property def backends()`.

jupyter_scheduler/backend_utils.py
```diff
@@ -0,0 +1,70 @@
...
+) -> str:
+    """Get backend ID for routing legacy jobs (UUID-only IDs from pre-3.0)."""
+    if not available_backends:
+        raise ValueError("No scheduler backends available.")
+
+    if legacy_job_backend and legacy_job_backend in available_backends:
+        return legacy_job_backend
+
+    if DEFAULT_FALLBACK_BACKEND_ID in available_backends:
+        return DEFAULT_FALLBACK_BACKEND_ID
```
dlqqq, line 65: `DEFAULT_FALLBACK_BACKEND_ID` is set to `jupyter_server_nb`. Is it possible for a legacy job to use a Python file instead of a notebook? It seems that would fail if `jupyter_server_nb` were used.

jupyter_scheduler/extension.py
```diff
@@ -45,6 +52,34 @@ class SchedulerApp(ExtensionApp):
...
 
+    legacy_job_backend = Unicode(
+        default_value=None,
+        allow_none=True,
+        config=True,
+        help=_i18n(
+            "Backend for jobs with UUID-only job IDs created before "
+            "multi-backend support (added in 3.0.0)."
+        ),
+    )
```
dlqqq, line 61: The comments throughout this PR imply that this PR would only be released in v3.0.

If we are considering a major release for this, I think there are a few things to consider:

1. It may be unnecessary to support legacy jobs. There is a lot of extra code being added to support this. If we include this logic with a v3.0 release, this would need to be maintained throughout all future v3 releases to follow semver.
2. We would also be able to remove some of the traitlets options that are difficult to adapt to this new backend, e.g. `scheduler_class`, which only works for legacy jobs IIUC.
3. We can upgrade from Pydantic v1 => Pydantic v2 to avoid using the `pydantic_v1` module.
4. We should do more refactoring and cut pre-releases before publishing the official release. I'm having difficulty understanding the multitude of components we have currently, and have a feeling that things can be simplified.
5. We would also want to create a `2.x` branch tracking latest `main` before merging to allow us to patch v2 if needed.

jupyter_scheduler/handlers.py
```diff
@@ -29,31 +34,68 @@
...
+
+async def call_async(scheduler, method: str, *args, **kwargs):
+    """Call scheduler method async. Uses native *_async method if available, else runs in thread pool."""
+    async_method = f"{method}_async"
+    if hasattr(scheduler, async_method):
+        logger.debug(f"Using native async method: {scheduler.__class__.__name__}.{async_method}")
+        return await getattr(scheduler, async_method)(*args, **kwargs)
+    logger.debug(f"Falling back to thread pool: {scheduler.__class__.__name__}.{method}")
+    return await asyncio.to_thread(getattr(scheduler, method), *args, **kwargs)
+
```
dlqqq: Note: This function can't be understood by type checking & static analysis tools and is therefore more fragile. If we're OK with breaking changes (see prev comment on v3.0 release), I would recommend making all methods `async` by default in the base classes.

jupyter_scheduler/orm.py
```diff
@@ -105,6 +105,7 @@ class Job(CommonColumns, Base):
     idempotency_token = Column(String(256))
     # All new columns added to this table must be nullable to ensure compatibility during database migrations.
     # Any default values specified for new columns will be ignored during the migration process.
+    backend = Column(String(64))
```
dlqqq: It's a little confusing that `job_id` in the DB means just the UUID, whereas `job_id` throughout the rest of the application means an ID of the format `{backend}:{uuid}`. I think that we should re-consider having this duplicate definition of what a "job ID" means.

jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+    module = __import__(module_path, fromlist=[class_name])
+    return getattr(module, class_name)
+
+
+@dataclass
+class BackendInstance:
+    """A running backend with its configuration and initialized scheduler."""
+
+    config: BackendConfig
+    scheduler: BaseScheduler
```
dlqqq: Since we are using `pydantic`, we don't need to use dataclasses.

dlqqq (commented):
@andrii-i Thanks for working on this over the last month! This is an impressive amount of code. 💪 

I left some code review feedback on the backend below. However, I haven't reviewed the backend or done local testing. Let's talk more about this PR Thursday (will miss standup tomorrow). Hopefully that will give @JGuinegagne time to review as well.

Regardless of we release this in v2 or v3, it would be great to publish a pre-release before publishing an official release. There are a lot of changes here, and I think we will need a bug bash to validate all of them.

andrii-i:
Thank you for the review David. Added "Fixes https://github.com/jupyter-server/jupyter-scheduler/issues/597, part of https://github.com/jupyter-server/jupyter-scheduler/issues/599" to the top of PR description to clarify that this is planned for release as a part of v3

jupyter_scheduler/tests/test_job_files_manager.py
```diff
@@ -162,3 +169,111 @@ def test_downloader_download(downloader_parameters):
...
+            mock_scheduler.get_job.return_value = job
+            mock_scheduler.get_staging_paths.return_value = staging_paths
+            mock_scheduler.get_local_output_path.return_value = output_dir
+            mock_scheduler.get_job_filenames.return_value = job_filenames
+
+            mock_backend = Mock()
+            mock_backend.scheduler = mock_scheduler
+
+            mock_registry = Mock()
+            mock_registry.get_backend.return_value = mock_backend
```
JGuinegagne: nit: I would define the mocks outside of the `with patch` context managers for better readability.

Also, you can combine patches:
```py
with patch(
    "jupyter_scheduler.job_files_manager.Downloader",
    "jupyter_scheduler.job_files_manager.Process"
) as mock_downloader, mock_process:
    pass
```

jupyter_scheduler/tests/test_job_id.py
```diff
@@ -0,0 +1,98 @@
+"""Tests for job ID encoding and parsing utilities."""
+
+from unittest.mock import MagicMock
+
+import pytest
+
+from jupyter_scheduler.job_id import make_job_id, parse_job_id, resolve_scheduler
+
+
+class TestMakeJobId:
```
JGuinegagne, line 10: I think you need `TestMakeJobId(unittest.TestCase)` here

jupyter_scheduler/tests/test_python_executor.py
```diff
@@ -0,0 +1,222 @@
...
+
+@pytest.fixture
+def python_script_staging_dir(jp_scheduler_staging_dir) -> Path:
+    """Create a staging directory with a simple Python script."""
+    job_staging_dir = jp_scheduler_staging_dir / "job-py-1"
+    job_staging_dir.mkdir()
+    return job_staging_dir
+
+
+@pytest.fixture
```
JGuinegagne, line 18: sanity-check: do you need [scope](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) for these fixtures?

Otherwise it defaults to `function`, which means in your case multiple mutate fs operations. It should be okay, but it's often frown upon in unit tests.

jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+        for cfg in self._configs:
+            try:
+                instance = self._create_backend(cfg, root_dir, environments_manager, db_url, config)
+                self._backends[cfg.id] = instance
+
+                for ext in cfg.file_extensions:
+                    ext_lower = ext.lower().lstrip(".")
+                    if ext_lower not in self._extension_map:
+                        self._extension_map[ext_lower] = []
+                    self._extension_map[ext_lower].append(cfg.id)
```
JGuinegagne, line 66: sanity-check: `cfg.id` are guaranteed to be mutually exclusive/unique, correct?

jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def _create_backend(
+        self,
+        cfg: BackendConfig,
+        root_dir: str,
+        environments_manager: EnvironmentManager,
+        global_db_url: str,
+        config: Optional[Any] = None,
+    ) -> BackendInstance:
+        """Create a backend instance from configuration."""
```
JGuinegagne: nit: non-descriptive docstring (repeat method name).

something like:
```py
"""Import scheduler class, instantiate scheduler and backend, return backend.

Create the database tables if not found.
"""
```

jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+            backend_id=cfg.id,
+        )
+
+        if cfg.execution_manager_class:
+            scheduler.execution_manager_class = import_class(cfg.execution_manager_class)
+
+        return BackendInstance(config=cfg, scheduler=scheduler)
+
+    def get_backend(self, backend_id: str) -> Optional[BackendInstance]:
+        """Get a backend by ID, or None if not found."""
```
JGuinegagne: `Return a backend with matching ID, None if none is found.`

jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+        return self._backends.get(backend_id)
+
+    def get_legacy_job_backend(self) -> BackendInstance:
+        """Get the backend for routing legacy jobs (UUID-only IDs from pre-3.0)."""
```
JGuinegagne: consider adding a `Raises: ` section to the docstring.

jupyter_scheduler/backend_registry.py
```diff
@@ -0,0 +1,149 @@
...
+        # 1. Explicit preference for this extension
+        preferred = self._preferred_backends.get(ext)
+        if preferred and preferred in candidates:
+            return self._backends[preferred]
+
+        # 2. Alphabetical by name
+        candidate_instances = [self._backends[bid] for bid in candidates]
+        return min(candidate_instances, key=lambda b: b.config.name)
+
+    def list_backends(self) -> List[DescribeBackend]:
```
JGuinegagne: sanity-check: how many backends could there be?

do we need to worry about payload size, and think about pagination?

jupyter_scheduler/backends.py
```diff
@@ -0,0 +1,74 @@
...
+    scheduler_class: str
+    execution_manager_class: str
+    database_manager_class: Optional[str] = None
+    db_url: Optional[str] = None
+    file_extensions: List[str] = field(default_factory=list)
+    output_formats: List[Dict[str, str]] = field(default_factory=list)
+    metadata: Optional[Dict[str, Any]] = None
+
+
+class DescribeBackend(BaseModel):
```
JGuinegagne: be explicit for API models: `DescribeBackendResponse`

jupyter_scheduler/exceptions.py
```diff
@@ -3,8 +3,6 @@ class SchedulerError(Exception):
     that are not defined here
     """
 
-    pass
```
JGuinegagne, line 6: sanity-check: okay to remove this?

jupyter_scheduler/extension.py
```diff
@@ -66,28 +104,89 @@ def _db_url_default(self):
         help=_i18n("The job files manager class to use."),
     )
 
+    def _build_backend_configs(self, backend_classes: dict) -> list:
```
JGuinegagne: optional: use better-defined arg & return types

jupyter_scheduler/extension.py
```diff
@@ -66,28 +104,89 @@ def _db_url_default(self):
...
-            scheduler=scheduler,
+            scheduler=scheduler,  # Backwards compatibility with handlers expecting single scheduler (uses legacy job backend)
+            backend_registry=registry,
             job_files_manager=job_files_manager,
         )
 
-        if scheduler.task_runner:
-            loop = asyncio.get_event_loop()
-            loop.create_task(scheduler.task_runner.start())
+        loop = asyncio.get_event_loop()
```
JGuinegagne, line 185: sanity-check: okay to remove this conditional? It...looks dangerous.

jupyter_scheduler/handlers.py
JGuinegagne: +1, any way to make the return type _not_ ambiguous?

jupyter_scheduler/handlers.py
```diff
@@ -29,31 +34,68 @@
...
+        return self._backend_registry
+
     @property
     def environments_manager(self):
         if self._environments_manager is None:
             self._environments_manager = self.settings.get("environments_manager")
-
         return self._environments_manager
 
+    def get_scheduler(self, job_id: str):
```
JGuinegagne: nit: add return type

jupyter_scheduler/handlers.py
```diff
@@ -29,31 +34,68 @@
...
+        except ValueError as e:
+            raise HTTPError(400, str(e))
+
+    def resolve_backend_for_job(self, payload: dict) -> BackendInstance:
+        """Resolve backend from payload['backend'] or auto-select by file extension."""
+        backend_id = payload.get("backend")
+        if backend_id:
+            backend = self.backend_registry.get_backend(backend_id)
+            if not backend:
+                raise HTTPError(400, f"Unknown backend: {backend_id}")
```
JGuinegagne: 404 ?

jupyter_scheduler/handlers.py
```diff
@@ -135,6 +181,9 @@ async def post(self):
         except SchedulerError as e:
             self.log.exception(e)
             raise HTTPError(500, str(e)) from e
```
JGuinegagne: sanity-check: I know this is old code, but do we want to return the error string for a 500?

jupyter_scheduler/handlers.py
```diff
@@ -213,7 +264,32 @@ async def get(self, job_id=None):
...
+                            try:
+                                synced_job = await call_async(
+                                    backend.scheduler, "get_job", job.job_id, job_files=False
+                                )
+                                list_jobs_response.jobs[i] = synced_job
+                            except Exception as e:
+                                self.log.warning(f"Failed to sync status for job {job.job_id}: {e}")
             except ValidationError as e:
                 self.log.exception(e)
                 raise HTTPError(500, str(e)) from e
```
JGuinegagne: same question.

JGuinegagne: Also, `ValidationError` map to 500, not 400 ?

jupyter_scheduler/handlers.py
```diff
@@ -415,3 +513,27 @@ async def get(self, job_id):
...
+        try:
+            registry = self.backend_registry
+            if registry is None:
+                raise HTTPError(500, "Backend registry not initialized")
+
+            backends = registry.list_backends()
+            self.finish(json.dumps([b.dict() for b in backends]))
+        except SchedulerError as e:
+            self.log.exception(e)
+            raise HTTPError(500, str(e)) from e
```
JGuinegagne: same question

jupyter_scheduler/job_files_manager.py
```diff
@@ -102,8 +129,8 @@ def download(self):
                     with fsspec.open(input_filepath) as input_file:
                         with fsspec.open(output_filepath, mode="wb") as output_file:
                             output_file.write(input_file.read())
-                except Exception as e:
-                    pass
+                except Exception:
+                    logger.error(f"Failed to download {input_filepath}")
```
JGuinegagne: you should capture the error message in the log

jupyter_scheduler/job_id.py
```diff
@@ -0,0 +1,29 @@
...
+def make_job_id(backend_id: str, uuid: str) -> str:
+    """Create a job ID in the standard 'backend_id:uuid' format."""
+    return f"{backend_id}:{uuid}"
+
+
+def parse_job_id(job_id: str) -> Tuple[Optional[str], str]:
+    """Parse job ID into (backend_id, uuid). Returns (None, uuid) for legacy IDs (no colon)."""
+    if ":" not in job_id:
+        return None, job_id
+    backend_id, uuid = job_id.split(":", 1)
```
JGuinegagne, line 17: sanity-check: `backend_id` are guaranteed not to contain `:`, correct?

jupyter_scheduler/orm.py
JGuinegagne: also, this value calls for an `_id` postfix IMHO

jupyter_scheduler/python_executor.py
```diff
@@ -0,0 +1,93 @@
...
+        if result.stderr:
+            with fsspec.open(stderr_path, "w", encoding="utf-8") as f:
+                f.write(result.stderr)
+
+        # Capture any additional side effect files AFTER writing stdout/stderr
+        self.add_side_effects_files(staging_dir)
+
+        if result.returncode != 0:
+            raise RuntimeError(
+                f"Script exited with code {result.returncode}\nstderr: {result.stderr[:500]}"
```
JGuinegagne: sanity-check: Full [:500] stderr trace in the error message?

jupyter_scheduler/python_executor.py
```diff
@@ -0,0 +1,93 @@
...
+            )
+
+    def add_side_effects_files(self, staging_dir: str):
+        """Scan for files created during execution and update job's packaged_files."""
+        input_script = os.path.basename(self.staging_paths["input"])
+        new_files = set()
```
JGuinegagne, line 55: typing please

jupyter_scheduler/python_executor.py
```diff
@@ -0,0 +1,93 @@
...
+            for key, value in job.parameters.items():
+                env[f"JUPYTER_PARAM_{key}"] = str(value)
+
+        result = subprocess.run(
+            [sys.executable, self.staging_paths["input"]],
+            cwd=staging_dir,
+            capture_output=True,
+            text=True,
+            env=env,
+        )
```
JGuinegagne, line 32: question: should we worry about timeouts here?

jupyter_scheduler/scheduler.py
```diff
@@ -96,11 +98,17 @@ def _default_staging_path(self):
     )
 
     def __init__(
-        self, root_dir: str, environments_manager: Type[EnvironmentManager], config=None, **kwargs
+        self,
+        root_dir: str,
+        environments_manager: Type[EnvironmentManager],
+        config=None,
```
JGuinegagne: typing?

jupyter_scheduler/scheduler.py
```diff
@@ -324,13 +332,16 @@ def add_job_files(self, model: DescribeJob):
         for output_format in model.output_formats:
             filename = output_filenames[output_format]
             output_path = os.path.join(output_dir, filename)
-            job_files.append(
-                JobFile(
-                    display_name=mapping[output_format],
-                    file_format=output_format,
-                    file_path=output_path if self.file_exists(output_path) else None,
+            file_exists = self.file_exists(output_path)
+            # Only add job file if it exists (handles optional outputs like stdout/stderr)
```
JGuinegagne, line 336: sanity-check: is it correct that if the file doesn't exists, then this job creation is bypassed in a silent no-op? shouldn't the user see an error instead?

src/mainviews/create-job.tsx
```diff
@@ -118,23 +123,65 @@ export function CreateJob(props: ICreateJobProps): JSX.Element {
...
+
+        const validBackends = filterBackendsByFile(
+          backends,
+          props.model.inputFile
+        );
+
+        // Only update if current backend is invalid for this file type
+        if (!validBackends.some(b => b.id === props.model.backend)) {
+          // Server returns backends sorted by preference; first is pre-selected
+          const selected = validBackends[0];
```
JGuinegagne: I'm a bit confused, line 154 doesn't guarantee that the first item as `b.id === props.model.backend`

Could you clarify the logic?

JGuinegagne (commented):
Lots of minor suggestions/sanity-checks.

Caveat: I'm not familiar with this codebase, some comments may be irrelevant or missing context. Feel free to use judgement on what to address.

jupyter_scheduler/backend_registry.py
andrii-i: Here `_preferred_backends` is `Dict[str, str]` of file extensions to backend IDs from `preferred_backends` traitlet, so `self._preferred_backends.get(ext)` returns a preferred backend ID (str). Only `_backends` stores `BackendInstance`s in `Dict[str, BackendInstance]`.

To improve clarity updated var naming in the function to mention ids, Updated helped text for `preferred_backends` and `legacy_job_backend` traitlets clearly explaining that they expect backend ids. 

Looking at recent traitlets docs it seems we could add typing https://traitlets.readthedocs.io/en/stable/using_traitlets.html#default-values-and-checking-type-and-value. Adding typing to all traitlets as currently we have none would be a good follow-up, created https://github.com/jupyter-server/jupyter-scheduler/issues/602 to track it.

jupyter_scheduler/backend_utils.py
andrii-i: Pre-v3.0 only supported / supports notebook `.ipynb` files and was not providing a system to allow support for more file formats so it was impossible for a legacy job to use a Python file instead of a notebook (unless someone would created another extension based on scheduler doing that).

jupyter_scheduler/extension.py
andrii-i: As discussed this is indeed to be released as a part of 3.0 https://github.com/jupyter-server/jupyter-scheduler/issues/599 

Re 1-2: dropping support for legacy jobs would definitely allow us to remove code and traitlet options needed to support them. It’s good to have backwards compatibility as it simplifies migration for existing users but we could absolutely look into this and consider tradeoffs again before the v3.

Re 3: I think v3 is definitely a great opportunity to upgrade to pydantic v2, created issue to track it and included it into the release plan / issue https://github.com/jupyter-server/jupyter-scheduler/issues/601.  

Re 4: absolutely open to optimization and refactoring where there’s an opportunity. Yes absolutely should / planned to do pre-releases

Re 5: also yes absolutely should / planned. 2.x created https://github.com/jupyter-server/jupyter-scheduler/tree/2.x

jupyter_scheduler/handlers.py
andrii-i: Removed `call_async` as I added it as a way to make list calls non-blocking quickly for a demo. I’ll make all methods async by default in the base classes in a separate PR as tracked in #598. This reverts code to using `ensure_async` for compatibility of sync scheduler methods with Tornado async handlers.

jupyter_scheduler/backend_registry.py
andrii-i: Agreed. It makes sense to pick one pattern instead of mixing both as it reduces cognitive load and makes the codebase easier to maintain, and we're already using Pydantic. Converted both `BackendConfig` and `BackendInstance` to Pydantic BaseModel.

jupyter_scheduler/orm.py
andrii-i: @dlqqq good point, changed logic so that `job_id` would always mean an ID of the format `{backend}:{uuid}` both in db and in application logic (unless it's a pre-3.0 legacy job, let's discuss need to support those separately). 

@JGuinegagne good point as well, renamed `backend` column to `backend_id` everywhere where we are referring to backend id (string) not backend class instance.

jupyter_scheduler/tests/test_python_executor.py
andrii-i: Good point. Extracted scripts to `jupyter_scheduler/tests/static/`, fixtures now just copy static files instead of creating them inline. Function scope is intentional since `jp_scheduler_staging_dir` creates a fresh temp directory per test for isolation. These are integration tests (testing actual script execution) so filesystem operations are expected.

jupyter_scheduler/tests/test_job_id.py
andrii-i: Current tests use pytest. Tests use plain assert statements and pytest.raises(), which work / are supposed to be used without TestCase. Adding TestCase inheritance won’t add anything and adding it would be mixing paradigms without benefit.

jupyter_scheduler/backend_registry.py
andrii-i: Backend IDs are author-defined strings so uniqueness is not guaranteed. Added duplicate check in BackendRegistry.initialize() that raises ValueError to prevent silent overwrite.

jupyter_scheduler/backend_registry.py
andrii-i: Added `Raises:` to this function and `get_legacy_job_backend` that also raises an error. Previously decided against adding the section for brevity / clarity from the code but after additional consideration I see additional UX value in also mentioning error logic in the docstring in compact readable manner.

jupyter_scheduler/backend_registry.py
andrii-i: We allow users to register any amount of supported file extensions and backends. At the same time each backend instantiates its own scheduler and potentially database connections, making it impractical to have too many due to overhead. I'd say typical deployments would have <10 backends at least initially.

The payload is relatively small - just ID, name, description, file_extensions, output_formats per backend. So combined with the fact that having too many backends is not practical due to overhead I'd say pagination would add complexity without practical benefit.

jupyter_scheduler/exceptions.py
andrii-i: Yes, no need for `pass` if there's a docstring which is a statement.

jupyter_scheduler/extension.py
andrii-i: It's not removed, see L186 below 
```
if hasattr(backend.scheduler, "task_runner") and backend.scheduler.task_runner:
```
This handles multiple backends and is safer against custom schedulers that might not define task_runner.

jupyter_scheduler/handlers.py
andrii-i: Makes sense.

jupyter_scheduler/handlers.py
andrii-i: I think having explicit error messages mentioning type of the operation during which error occurred seems like a clear improvement. Updated in all places where we raised 500 without the error string.

jupyter_scheduler/handlers.py
andrii-i: Good catch. `ValidationError` overall and in our codebase indicates the client sent invalid data that doesn't match the expected schema which should be 400 Bad Request (client error), not a 500 (server error). Updated all handlers.

jupyter_scheduler/job_files_manager.py
andrii-i: Good point. Changed to `logger.exception()` which includes the full traceback.

jupyter_scheduler/job_id.py
andrii-i: Now they are thanks to validation during registration, also added a test about it. Good catch.

src/mainviews/create-job.tsx
andrii-i: This handles auto-selecting a valid backend when the user opens Create Job for a file type but the model has a preferred backend that doesn't support it. Simplified with a named variable for readability and added a comment explaining the logic better.

jupyter_scheduler/python_executor.py
andrii-i: Good point. Changed to just error code and reference to the 'Errors' output for full error trace.

jupyter_scheduler/python_executor.py
andrii-i: Yes but let's consider it out of scope of this PR. Let's use this existing issue to track need to implement timeouts: https://github.com/jupyter-server/jupyter-scheduler/issues/90.

jupyter_scheduler/scheduler.py
andrii-i: No. `add_job_files()` populates the output files list after job completion. Each backend registers output formats and the check correctly adds files that actually exist to the model. For example, if a Python script produces no stdout, there's no stdout file to show even though `JupyterServerPythonBackend` registers it.

jupyter_scheduler/tests/utils.py
```diff
@@ -1,28 +1 @@
...
-        ]
-    ):
-        if expected_code != e.code:
-            return False
-        if expected_message:
-            message = json.loads(e.response.body.decode())["message"]
-            if expected_message != message:
-                return False
-        return True
+# Test utilities module (currently empty - add helpers here as needed)
```
JGuinegagne, line 1: non-blocking: remove module?

jupyter_scheduler/backend_registry.py
```diff
@@ -17,12 +16,11 @@ def import_class(class_path: str) -> Type:
...
 
 
-@dataclass
-class BackendInstance:
+class BackendInstance(BaseModel):
     """A running backend with its configuration and initialized scheduler."""
 
     config: BackendConfig
-    scheduler: BaseScheduler
+    scheduler: Any  # BaseScheduler at runtime, but Any to support test mocks
```
JGuinegagne, line 23: hum, compromising typing integrity for the sake of unit tests doesn't sound right...

jupyter_scheduler/backend_registry.py
```diff
@@ -48,6 +46,14 @@ def initialize(
         config: Optional[Any] = None,
     ):
         """Instantiate all backends from configs."""
+        seen_ids = set()
+        for cfg in self._configs:
+            if cfg.id in seen_ids:
+                raise ValueError(f"Duplicate backend ID: '{cfg.id}'")
+            if ":" in cfg.id:
+                raise ValueError(f"Backend ID cannot contain ':': '{cfg.id}'")
```
JGuinegagne, line 54: optional: performing two validation operations in the same loop, this is a bit odd.
Any way we could use `pydantic` models to validate the ID regex?

`pydantic` most likely supports disallowing duplicates in list.

jupyter_scheduler/handlers.py
```diff
@@ -71,20 +61,20 @@ def environments_manager(self):
             self._environments_manager = self.settings.get("environments_manager")
         return self._environments_manager
 
-    def get_scheduler(self, job_id: str):
+    def get_scheduler(self, job_id: str) -> BaseScheduler:
         """Get scheduler for a job ID. Raises HTTPError(400) if backend unavailable."""
```
JGuinegagne, line 65: non-blocking: I know it's not from this PR, but docstring formalities would call for:
```py
"""Return scheduler for a job ID.

Raises:
    HTTPError(400) if backend unavailable.
```

jupyter_scheduler/handlers.py
```diff
@@ -71,20 +61,20 @@ def environments_manager(self):
...
         try:
             return resolve_scheduler(job_id, self.backend_registry)
         except ValueError as e:
             raise HTTPError(400, str(e))
 
     def resolve_backend_for_job(self, payload: dict) -> BackendInstance:
-        """Resolve backend from payload['backend'] or auto-select by file extension."""
-        backend_id = payload.get("backend")
+        """Resolve backend from payload['backend_id'] or auto-select by file extension."""
+        backend_id = payload.get("backend_id")
```
JGuinegagne, line 73: optional: can `backend_id` possibly be nullish here?

jupyter_scheduler/handlers.py
```diff
@@ -120,16 +110,18 @@ class JobDefinitionHandler(ExtensionHandlerMixin, JobHandlersMixin, APIHandler):
...
-                    self.scheduler, "get_job_definition", job_definition_id
+                job_definition = await ensure_async(
+                    self.scheduler.get_job_definition(job_definition_id)
                 )
             except SchedulerError as e:
                 self.log.exception(e)
-                raise HTTPError(500, str(e)) from e
+                raise HTTPError(
+                    500, f"Unexpected error while getting job definition details: {e}"
+                ) from e
```
JGuinegagne, line 120: sanity-check: generally displaying the error string on a 5xx is frown-upon.
You're already logging it on line 117. Might be okay in your context though, so apply judgement.

jupyter_scheduler/handlers.py
```diff
@@ -307,39 +302,38 @@ async def post(self):
...
             self.log.exception(e)
-            raise HTTPError(500, str(e)) from e
+            raise HTTPError(500, f"Unexpected error during creation of job: {e}") from e
         except HTTPError:
             # Re-raise HTTPError as-is (e.g., 400 for invalid backend)
             raise
         except Exception as e:
             self.log.exception(e)
-            raise HTTPError(500, "Unexpected error occurred during creation of job.") from e
+            raise HTTPError(500, f"Unexpected error during creation of job: {e}") from e
```
JGuinegagne, line 332: optional: per DRY principle, consider exploring context manager for error handling

jupyter_scheduler/python_executor.py
```diff
@@ -46,10 +46,10 @@ def execute(self):
 
         if result.returncode != 0:
             raise RuntimeError(
-                f"Script exited with code {result.returncode}\nstderr: {result.stderr[:500]}"
+                f"Script exited with code {result.returncode}. See 'Errors' output for full error trace."
```
JGuinegagne, line 49: in that case, it would help to provide the file path.

ui-tests/tests/jupyter_scheduler.spec.ts
```diff
@@ -50,7 +50,8 @@ test.describe('Jupyter Scheduler', () => {
     await scheduler.createJobTbutton.click();
     await page.waitForSelector('text=Loading …', { state: 'hidden' });
     await page.waitForSelector('text=Saving Completed', { state: 'hidden' });
-    await scheduler.assertSnapshot(FILENAMES.CREATE_JOB_VIEW);
+    // Flaky: file names and timestamps vary by environment
+    // await scheduler.assertSnapshot(FILENAMES.CREATE_JOB_VIEW);
```
JGuinegagne, line 54: remove?

JGuinegagne (approved):
LGTM w/ optional comments.

Evaluate whether to return error details for 5xx (disallowed in a normal service, but may be okay in a client application like jupyter-scheduler).

dlqqq (approved):
Thank you for addressing our feedback Andrii!

I'm approving this PR because the main branch will continue to evolve, and I don't think see any issues worth blocking on right now. The testing you've added inspires confidence that this new feature works, so I think it's fine to merge for now.

Before the v3.0 release, we should revisit the architecture as a team, remove any unnecessary / excessively complex components, and simplify as much as possible to minimize maintenance burden. We should also think more deeply about schema changes for the local scheduler database, and whether we should add a DB migration script for users.

andrii-i:
Thanks for the review and approval @dlqqq. Would be happy to have additional discussions.

> whether we should add a DB migration script for users

The current [`update_db_schema function`](https://github.com/jupyter-server/jupyter-scheduler/blob/9d5e2ea50f5a9b1746c214fd7e4bb5f87fb768c8/jupyter_scheduler/orm.py#L123) handles automatic column additions via ALTER TABLE - new nullable columns are added transparently on startup. That said, happy to discuss if we need something more robust.

jupyter_scheduler/handlers.py
andrii-i: Generally yes.

But JupyterLab grants users full code execution and filesystem access, so anyone with access already has more privileges than error details would provide. Detailed errors help users debug without introducing additional security concerns.

Open to reconsidering due to evolution of how and where JupyterLab is deployed in a separate PR.

PR #609, opened 2026-03-04
Remove `fsspec` version ceiling, set to `>=2025.3.2`

andrii-i:
## Problem

Downstream consumers cannot upgrade to newer versions of `fsspec` (>= 2026.2.0) because `jupyter-scheduler` pins `fsspec>=2023.6.0,<=2025.3.2,!=2025.3.1`.

## Proposed solution

Remove the upper bound on `fsspec` and set the minimum to `>=2025.3.2`.

This would allow existing users to upgrade jupyter-scheduler without being forced to change their environment, while also unblocking users that need fsspec 2026.2.0.

## Code changes

Changed `fsspec>=2025.3.2`.

## Testing

On top of CI, verified manually that notebook job creation, execution, and output file download work correctly with `fsspec 2026.2.0`.

## User-facing changes

None.

The minimum required `fsspec` version is now `2025.3.2` (previously `2023.6.0`). Environments with older `fsspec` versions will need to upgrade.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "sqlalchemy>=2.0,<3",
     "croniter~=1.4",
     "pytz>=2023.3,<=2024.2",
-    "fsspec>=2023.6.0,<=2025.3.2,!=2025.3.1",
+    "fsspec>=2026.2.0",
     "psutil~=5.9"
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ version = "2.11.0"
 description = "A JupyterLab extension for running jobs"
 readme = "README.md"
 license = { file = "LICENSE" }
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 authors = [
     { name = "Project Jupyter" },
 ]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -36,7 +35,7 @@ dependencies = [
     "sqlalchemy>=2.0,<3",
     "croniter~=1.4",
     "pytz>=2023.3,<=2024.2",
-    "fsspec>=2023.6.0,<=2025.3.2,!=2025.3.1",
+    "fsspec>=2026.2.0",
```
dlqqq: Agree with removing the version ceiling. That has caused too many environment issues in the past.

However, can we do `fsspec>=2025.3.2` to have some overlap with the previous version range?

dlqqq (commented):
Looks great, thanks! Just one minor comment

pyproject.toml
dlqqq: Done in latest commit, thanks!

dlqqq (approved):

pyproject.toml
andrii-i: Yes, I think it's a good idea. Would allow existing users to upgrade `jupyter-scheduler` without being forced to change their environment, while also unblocking users that need `fsspec 2026.2.0`.

